PR #112, opened 2014-06-20
Do not claim annotations in the processors.

JakeWharton:


akanter:
@eamonnmcmanus @cgruber As far as errors go, if javac is passed the `-Xlint:processing` flag, then this will cause  a warning like
```
[WARNING] COMPILATION WARNING :
[INFO] -------------------------------------------------------------
[WARNING] No processor claimed any of these annotations: 
com.google.auto.value.AutoValue
```
The compiler will also spit out the rest of the annotations that you may be using that don't have processors (e.g., Jackson JSON annotations).
In this case, if you also have `-Werror` set on javac then your builds will fail.

I understand your logic in not wanting to claim the annotation, but this information should probably live somewhere in the README.

(To those who are interested in the workaround, either remove the `-Xlint:processing` flag, or if `-Xlint:all` is set then add the `-Xlint:-processing` flag.)

eamonnmcmanus:
This warning is a horrible javac misfeature and it's annoying to have to waste any time catering to it. You're typically going to see warnings for things like `@Nullable` or `@VisibleForTesting` or whatever, in addition to the Jackson annotations you mention.

You're probably right that we should tell people what to do if they encounter the problem, though.

PR #126, opened 2014-08-16
added intellij project files to `.gitignore`.

jmcampanini:
added a couple of intellij project file extensions to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,9 +17,12 @@ release.properties
 
 .idea
 *.iml
+*.ipr
+*.iws
 classes
 
 obj
 
 .DS_Store
 *~
+


tbroyer:
My 2c.: all those dev-env-specific excludes have nothing to do in a project's gitignore. If you use IntelliJ IDEA, setup a global .gitignore that ignores the IDEA files for all your projects.

cf. https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

nglorioso:
Sorry it took so long, but in the intervening time, it looks like IntelliJ migrated to the `.idea`-based project files, so these changes aren't needed anymore.

PR #145, opened 2014-10-16
Fix top level class factory generation

danielnorberg:
FactoryWriter.getPackage() assumed that the fully qualified class name would contain a package and hence failed with a StringIndexOutOfBoundsException for top-level classes without a package.


netdpb:
I think this was fixed by 12436ccd31725220fc90fe3775a58e39930508b5.

PR #152, opened 2014-10-28
Some maven users may wonder why just adding optional dep doesn't work.

adriancole:
Potential documentation nudge might help. This came up in a downstream project.

https://github.com/jclouds/jclouds/pull/590


cgruber:
So this is odd, and seriously just a maven configuration fail on their part - they overrode the defaults which collect generated resources in target/classes, without specifying that in the new specified set.  I'm not sure why they're even configuring resources, since they're configuring (effectively) the defaults (in jclouds).  The instructions you put here shouldn't be necessary unless someone is already mucking with maven defaults.  Is it worth having in our isntructions when the vast majority of cases shouldn't need this?

Maybe we can have some sort of troubleshooting page, and have it there... but I suspect people will just cargo-cult this into their builds and it will actually be wrong if they aren't also specifying all the resource locations.

tbroyer:
The culprit is the maven-bundle-plugin, which by defaults only includes resources from the Maven Project model, and not those that have been processed already by the maven-resources-plugin.

> The `<Include-Resource>` instruction is a list of arbitrary resources that should be copied into the bundle JAR file.
> […]
> By default the bundle plugin converts the project's Maven resource directories into a single `<Include-Resource>` instruction.
— Source: https://felix.apache.org/site/apache-felix-maven-bundle-plugin-bnd.html#ApacheFelixMavenBundlePlugin%28BND%29-%7B%7B%3CIncludeResource%3E%7D%7D

A few lines further below, we can read that “`<Include-Resource>` is generated from the project's Maven resources, typically "`src/main/resources/`", which will copy the specified project directory hierarchy into the resulting bundle JAR file, **mirroring standard Maven behavior.**” (emphasis mine; reading this, it looks like it doesn't filters the resources, [looking at the sources](https://github.com/apache/felix/blob/c4dfed375642f019919822ebde1b1a00491db900/bundleplugin/src/main/java/org/apache/felix/bundleplugin/BundlePlugin.java#L1534-L1538), it appears BND will do the filtering).

The real fix for jclouds would be to configure the maven-bundle-plugin with `<Include-Resource>{maven-resources},META-INF/services/**=${project.build.outputDirectory}/META-INF/services/**</Include-Resource>` (well, the real fix would be to fix the maven-bundle-plugin to make it a better Maven "citizen")

tbroyer:
See https://felix.apache.org/site/apache-felix-bundle-plugin-faq.html#ApacheFelixBundlePluginFAQ-Whydosomegeneratedresources%2528suchasDeclarativeServicesXMLfiles%2529appearinthefinaljar%252Cbutothersdon%2527t%253F

adriancole:
I'm inclined to close this. Folks who have a similar issue will find this thread in an issues search. Thanks for the investigation! cc @demobox

demobox:
Thanks for the ping, @adriancole and the investigation, @tbroyer! I'll try to put together a PR for jclouds to fix this.

Issue #156, opened 2014-11-10
AutoValue: asm dependency should be shadowed


jbgi:
version conflict can occur (and do on one of my projects). It would be good if auto-value could shade all of it dependencies, eg. with maven-shade-plugin, like in https://github.com/immutables/immutables/blob/master/value-standalone/pom.xml
I could run auto-value annotation processing separately from compile, but I'd rather avoid the hassle.

PR #160, opened 2014-11-18
Change the AutoValue pom.xml so that it uses the Maven shade plugin to i...

eamonnmcmanus:
Change the AutoValue pom.xml so that it uses the Maven shade plugin to include all dependencies in the generated jar, and to rename most of them. This is intended to address https://github.com/google/auto/issues/156.

### DIFF
--- a/value/pom.xml
+++ b/value/pom.xml
@@ -123,6 +123,46 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-shade-plugin</artifactId>
+	<version>2.3</version>
+	<executions>
+	  <execution>
+	    <phase>package</phase>
+	    <goals>
+	      <goal>shade</goal>
+	    </goals>
+	    <configuration>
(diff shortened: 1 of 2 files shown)


jmax01:
It would be preferable to release a shaded version and one with explicit dependencies.
I spent a couple hours tracking down a bug that resulted from an issue in auto-value's commons-lang version, 2.4, which was addressed in 2.5.

eamonnmcmanus:
Sorry to hear about your wasted time, @jmax01. Does the bug in question manifest itself in AutoValue's behaviour or is it just that something else was picking up commons-lang from the AutoValue jar? If the latter, there's no real reason for that jar to be in the classpath at runtime, so you probably want to get rid of it if you can. Possibly the documentation could be clearer on this issue.

Also, it wouldn't be an issue if we could shade commons-lang. I didn't do this initially because of some reflection stuff that it does, but it might be that we could work around that.

jmax01:
*is it just that something else was picking up commons-lang from the AutoValue jar*
This was the problem. A dependency used HashcodeBuilder, which is broken in 2.4.

PR #175, opened 2015-01-21
Merge from google internal repository

cgruber:
Changes include, primarily:

  * AutoValue builders
  * AnnotationValues, AnnotationMirrors utilities in common
  * BasicAnnotationProcessor in common with validation and multi-round support for types which may not exist until those later rounds.
  * Various fixes and improvements to auto common, notably around handling generics and union types.


cgruber:
@eamonnmcmanus PTAL

cgruber:
Manually removed false-negative CLA tag, as all commit authors are googlers.

common/src/main/java/com/google/auto/common/MoreTypes.java
```diff
@@ -205,9 +243,40 @@ private static boolean equal(TypeMirror a, TypeMirror b, Set<ComparedElements> v
     EqualVisitorParam p = new EqualVisitorParam();
     p.type = b;
     p.visiting = visiting;
+    if (INTERSECTION_TYPE != null) {
+      if (INTERSECTION_TYPE.isInstance(a)) {
```
JakeWharton: On JavaWriter @tbroyer pointed out that the docs for `TypeMirror` discourage this.

> To implement operations based on the class of an TypeMirror object, either use a visitor or use the result of the getKind() method. Using instanceof is not necessarily a reliable idiom for determining the effective class of an object in this modeling hierarchy since an implementation may choose to have a single object implement multiple TypeMirror subinterfaces.

common/src/main/java/com/google/auto/common/MoreTypes.java
```diff
@@ -205,9 +243,40 @@ private static boolean equal(TypeMirror a, TypeMirror b, Set<ComparedElements> v
...
+  // between Java 7 and Java 8. In Java 7 it was modeled as a fake DeclaredType, and our logic
+  // for DeclaredType does the right thing. In Java 8 it is modeled as a new type IntersectionType.
+  // In order for our code to run on Java 7 (and Java 6) we can't even mention IntersectionType,
+  // so we can't override visitIntersectionType(IntersectionType). Instead, we discover through
+  // reflection whether IntersectionType exists, and if it does we extract the bounds of the
+  // intersection ((Number, Comparable<T>) in the example) and compare them directly.
+  @SuppressWarnings("unchecked")
+  private static boolean equalIntersectionTypes(
+      TypeMirror a, TypeMirror b, Set<ComparedElements> visiting) {
+    if (!INTERSECTION_TYPE.isInstance(b)) {
```
JakeWharton: Same thing here.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
```diff
@@ -185,4 +185,105 @@ final class $subclass$formalTypes extends $origClass$actualTypes {
...
+  #foreach ($m in $toBuilderMethods)
+
+  @Override
+  public ${builderTypeName}${builderActualTypes} ${m}() {
+    return new Builder${builderActualTypes}(this);
+  }
+
+  #end
+
+  static final class Builder${builderFormalTypes} ##
```
JakeWharton: +`private`? This type isn't used anywhere but from `builder()`. Or do you allow people to reference the type and its constructor directly?

value/src/main/java/com/google/auto/value/processor/autovalue.vm
```diff
@@ -185,4 +185,105 @@ final class $subclass$formalTypes extends $origClass$actualTypes {
...
+
+    private final $bitSet set$ = new ${bitSet}(${props.size()});
+
+    #foreach ($p in $props)
+
+    private $p.type $p;
+
+    #end
+
+    Builder() {
```
JakeWharton: Also +`private`? Same as https://github.com/google/auto/pull/175/files#r23339147

common/src/main/java/com/google/auto/common/MoreTypes.java
JakeWharton: https://github.com/square/javapoet/pull/93#discussion-diff-21441884

eamonnmcmanus: Hmm, that is a good point. Indeed there are other methods in this class whose entire raison d'être is to handle the theoretical case where a TypeMirror might be instanceof something without actually representing that something. I will make a later change that uses a.getKind().name() instead. That will allow us to continue to work on pre-Java-8 without requiring the introduction of yet more reflection objects.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
eamonnmcmanus: I'm not sure if you're referring to an earlier version where we would define a static builder() method in AutoValue_Foo? We don't do that any more: the Foo class is expected to do new AutoValue_Foo.Builder() directly.

eamonnmcmanus:
This doesn't build, because of a reference to the `@GoogleInternal` method `FluentIterable.getOnlyElement`.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
JakeWharton: Yep. Totally my mistake!

cgruber:
Fixed, though it's also relying on AutoValueBuilderProcessor, which didn't get the @GoogleInternal removed as intended.  Given that AutoValue is failing, but I don't want to do the entire re-sync with Moe from scratch, with these fixes too, can we just check this in as-is, and un-internalize AutoValueBuilderProcessor in a new CL,and I'll fix this up post-haste with a new moe run?

eamonnmcmanus:
Sounds like a plan.

cgruber:
Are you going to do CLs addressing Jake's comments as well?

eamonnmcmanus:
OK.

value/src/main/java/com/google/auto/value/processor/AnnotationOutput.java
```diff
@@ -0,0 +1,244 @@
...
+      break;
+    case '\r':
+      sb.append("\\r");
+      break;
+    case '\t':
+      sb.append("\\t");
+      break;
+    default:
+      if (c < 0x20) {
+        sb.append(String.format("\\%03o", (int) c));
```
swankjesse: ew! octal escapes?

eamonnmcmanus: I guess \u escapes would work too, since we've already eliminated the characters (\n and \r) that could cause problems, given that \u escapes are replaced before string constants are parsed.

PR #178, opened 2015-01-22
Add missing explicit dependency on Truth for AutoValue's functional test...

eamonnmcmanus:
...s.


cgruber:
Still failing - check Travis - seems like something with the builder-with-generics functional test. Though I'll merge this (as it is correct) and we can chat internally about that last error.

PR #180, opened 2015-01-31
AutoService is not necessarily should leave traces in class files!

SaltOfTheFlame:


JakeWharton:
:+1: This was fixed for other autos. Guess this guy was missed.

cgruber:
Can you please update the CLA per the instructions?

SaltOfTheFlame:
what shoulld I do with that thing? I've already agreed to CLA

ronshapiro:
In the meantime, we had done something similar. I'll close this.

Issue #173, opened 2015-01-14
AutoFactory depends on Dagger 1, compilation fails with Dagger 2


cypressious:
When building with AutoFactory and Dagger 2, the build fails with the message ```java.lang.NoClassDefFoundError: dagger/ObjectGraph```. In my build I specifically excluded the Dagger 1 dependency from AutoFactory. Here's an extract from my build configuration:

```Groovy
dependencies {
    compile 'com.google.dagger:dagger:2.0-SNAPSHOT'
    compile 'javax.annotation:jsr250-api:1.0'
    compile ('com.google.auto.factory:auto-factory:0.1-beta1') {
        exclude module: 'dagger'
    }
    apt ('com.google.auto.factory:auto-factory:0.1-beta1') {
        exclude module: 'dagger'
    }
    apt 'com.google.dagger:dagger-compiler:2.0-SNAPSHOT'
    provided 'org.glassfish:javax.annotation:10.0-b28'
}
```

I don't mind AutoFactory depending on Dagger 1 as long as it doesn't interfer with my build. Maybe the Dagger 1 dependency could be repackaged?

cgruber:
#230 removes use of dagger in any capacity within the processor itself, leaving only support for JSR-330 usage in client code.

Issue #111, opened 2014-06-20
AutoBuilder?


ooxi:
Hi,

is there something like an `AutoBuilder` planned? Looking at the following class

```
@AutoValue
public abstract class MyValue {
    public abstract String getFirst();
    public abstract String getSecond();
}
```

it would be great if I could do the following

```
MyValue value = MyValue.builder()
    .withFirst("hello")
    .withSecond("world")
.build();
```

Or are there compelling reasons against? I know of auto factories but they serve another purpose.

-- ooxi

felipecsl:
Looks like this has been just added a few days ago https://github.com/google/auto/pull/175

ooxi:
@felipecsl Looks great, is there any documentation on how to use this?

tbroyer:
@ooxi Have a look at the javadoc, there's a new `@AutoValue.Builder` annotation.

eamonnmcmanus:
Yes, there are a few final improvements and a documentation change in the pipeline and they should show up at the next sync from Google's internal branch. Feel free to experiment; the javadoc should be enough to get the gist.

ooxi:
:+1: great improvement :)

h908714124:
AutoBuilder? Yes! https://github.com/h908714124/auto-builder

Issue #99, opened 2014-05-14
AutoFactory: factories are final, prevent use of Mockito


cgrushko:
Generated factories are final, which means they cannot be mocked using Mockito. 
Even though they're designed for injection, passing a mock of such factory is so much easier than providing a test module and messing around with the DI's graph.

cjp39:
What was the result of this discussion at Google?

If this issue is still open for discussion, I'd like to note that declaring classes **final** was only listed as the _easier_ approach to prohibiting extension in Effective Java; making constructors private is also valid. In this case, the unfortunate side-effect of declaring this generated API final is that you discourage people from using `AutoFactory` when they find it's final. (Real-life examples are available on request.) The work-around is not obvious, and adds many lines of unnecessary boiler-plate that needs to be kept in-sync with other code. Unfortunately it seems like Daggers † and ‡ don't support `@Inject`-annotated static factory methods, which presumably would mean _they_ would need changes, too (and I have no idea how controversial that would be).

If someone's able to clarify the state of all this, that'd be fab.

cjp39:
Oh, and I've just realised there's an `allowSubclasses` property on `@AutoFactory` that presumably removes the `final` keyword. So, I guess this can be closed? Unless there was a plan to use `@Inject`-annotated static factory methods even when `allowSubclasses` was false?

tbroyer:
This was fixed in e0e2394ec33cb3d58e39fa142f138fefd517b1a3 BTW; and it's properly documented in the javadoc https://static.javadoc.io/com.google.auto.factory/auto-factory/1.0-beta3/com/google/auto/factory/AutoFactory.html#allowSubclasses-- ;-)

emartynov:
Do I understand correctly, that I want to mock factories I should use `allowSubclasses = true`?

PR #183, opened 2015-02-03
Synchronize and merge in google internal changes.

cgruber:
Notably:
  * AutoFactory simple factories can be configured to generate non-final factories for subclassing and mocking (addresses #99)
  * Updates AutoFactory to use Dagger 2 internally (should address #173)
  * Updates documentation for AutoValue to describe the AutoValue.Builder support (relates to #111)
  * Improve `@Nullable` support in AutoValue builders. 
  * Some improvements to common's MoreTypes with respect to parameterized and intersection types, and element conversion


value/README.md
```diff
@@ -201,9 +246,29 @@ You can offer as many static creation methods as you need, named
...
 They do not need to be named `create` as in the example.
 
+### Default values with builders
+Generated builders require every property to be set, except
+`@Nullable` properties. If you want to define a default value for a
+property, set it in the `builder()` method before returning. Here's
+how to define that animals have 4 legs by default:
+
+```java
+    class Example {
```
JakeWharton: All these code blocks don't need indented.

eamonnmcmanus: Perhaps, but that would be a subject for another PR.

eamonnmcmanus:
Do we know what the AutoFactory compilation failure in Travis means?

value/src/main/java/com/google/auto/value/processor/autovalue.vm
```diff
@@ -201,7 +204,7 @@ final class $subclass$formalTypes extends $origClass$actualTypes {
   #if ($builderIsInterface) implements #else extends #end
       ${builderTypeName}${builderActualTypes} {
 
-    private final $bitSet set$ = new ${bitSet}(${props.size()});
+    private final $bitSet set$ = new ${bitSet}();
```
JakeWharton: What's the motivation in removing the count initialization rather than using the non-nullable property count?

cgruber:
Wow.  I have no idea.  It built on my machine before I pushed it.  That's... odd.  Let me dig deeper.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
eamonnmcmanus: The non-nullable property count is not readily available, and the default BitSet size is 64 bits, so it would only make a difference if there were more than 64 properties anyway.

PR #192, opened 2015-02-19
Remove the use of the obsolescent com.google.common.base.Charsets from some tests.

cgruber:
Remove the use of the obsolescent com.google.common.base.Charsets from a couple of AutoValue tests. It's trivial to replace its use with Charset.forName.

Also change NoVelocityLoggingTest so that it also uses Charset.forName instead of using java.nio.charset.StandardCharsets, since the latter is unavailable on Java 6. (Apparently we haven't been running tests on Java 6, so who knows what else might be broken.)
-------------
Created by MOE: http://code.google.com/p/moe-java
MOE_MIGRATED_REVID=86527338

### DIFF
--- a/value/src/test/java/com/google/auto/value/processor/CompilationErrorsTest.java
+++ b/value/src/test/java/com/google/auto/value/processor/CompilationErrorsTest.java
@@ -15,7 +15,6 @@
  */
 package com.google.auto.value.processor;
 
-import com.google.common.base.Charsets;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ArrayListMultimap;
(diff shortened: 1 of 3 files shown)


JakeWharton:
Even if you target Java 6 why would you bother with JDK 6 to build? It's been EOL for a long time.

eamonnmcmanus:
I don't understand the question. If we want to test that AutoValue works on Java 6, then the tests should be built on Java 6.

JakeWharton:
The question was, why bother? Java 6 is a long-dead man.

jbgi:
Some big company IT departments excel in technological necromancy.

PR #194, opened 2015-02-24
Fix to import order

cgruber:


cgruber:
Minor stuff in google's internal repo, likely altered by IDE settings automatically updating things.  Pushing out so we don't have superfluous merge conflicts.

common/src/test/java/com/google/auto/common/SuperficialValidationTest.java
```diff
@@ -274,7 +274,7 @@ public void invalidAnnotationValue() {
   }
   */
 
-  private abstract static class AssertingProcessor extends AbstractProcessor {
```
eamonnmcmanus: Google style and the [JLS](http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1.1) both say that this is already the correct order. Where is the other one coming from?

cgruber:
FYI, @eamonnmcmanus - these are two no-op changes.  I'm going to merge them pre-LGTM (TBR essentially) so I can get a MOE equivalence.

cgruber:
Oh, NVM.  You're doing your thing.

common/src/test/java/com/google/auto/common/SuperficialValidationTest.java
cgruber: Um... dunno - but it got flipped internally.  Maybe we need to un-flip it.  :/  I'll revert this one change, and we can push it the other direction then.

cgruber:
Ok - it's only the import order now.

PR #195, opened 2015-02-24
Add a local tmp-file based repository for build-only pom.

cgruber:
Add a local tmp-file based repository for distribution of the build-only pom, so that a 'mvn deploy' command applied to it will simply put its artifact harmlessly into a local temp directory.

This is not inherited by any other project, since they do not list the build-pom as a parent, so it is purely to tuck away the build convenience pom when doing whole-project builds.

### DIFF
--- a/build-pom.xml
+++ b/build-pom.xml
@@ -12,4 +12,16 @@
     <module>service</module>
     <module>value</module>
   </modules>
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
LGTM

PR #196, opened 2015-02-24
Auto-push snapshot releases and javadocs. 

cgruber:
Add travis configuration, secure variables, and scripts suitable to add automatic snapshot and javadoc publishing on a successful build.

### DIFF
--- a/util/generate-latest-docs.sh
+++ b/util/generate-latest-docs.sh
@@ -1,0 +1,25 @@
+# see http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/ for details
+
+if [ "$TRAVIS_REPO_SLUG" == "google/truth" ] && \
+   [ "$TRAVIS_JDK_VERSION" == "oraclejdk7" ] && \
+   [ "$TRAVIS_PULL_REQUEST" == "false" ] && \
+   [ "$TRAVIS_BRANCH" == "master" ]; then
+  echo -e "Publishing javadoc...\n"
+  
+  mvn -f build-pom.xml javadoc:aggregate
+  TARGET="$(pwd)/target"
+
+  cd $HOME
+  git clone --quiet --branch=gh-pages https://${GH_TOKEN}@github.com/google/truth gh-pages > /dev/null
+  
+  cd gh-pages
+  git config --global user.email "travis@travis-ci.org"
+  git config --global user.name "travis-ci"
+  git rm -rf api/latest 
+  mv ${TARGET}/site/apidocs api/latest
+  git add -A -f api/latest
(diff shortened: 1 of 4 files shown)


util/generate-latest-docs.sh
```diff
@@ -0,0 +1,25 @@
...
+   [ "$TRAVIS_JDK_VERSION" == "oraclejdk7" ] && \
+   [ "$TRAVIS_PULL_REQUEST" == "false" ] && \
+   [ "$TRAVIS_BRANCH" == "master" ]; then
+  echo -e "Publishing javadoc...\n"
+  
+  mvn -f build-pom.xml javadoc:aggregate
+  TARGET="$(pwd)/target"
+
+  cd $HOME
+  git clone --quiet --branch=gh-pages https://${GH_TOKEN}@github.com/google/truth gh-pages > /dev/null
```
eamonnmcmanus: Not sure I fully understand the environment this is running in. Is $HOME empty each time? So no gh-pages directory already there?

cgruber: I believe that is the case.

cgruber: Admittedly, there is a bit of cargo-culting going on here, but this is the same pattern used in all of our other projects that do this.

sameb: URL is wrong.  should be auto, not truth?  same w/ the repo slug check at the top.

util/generate-latest-docs.sh
```diff
@@ -0,0 +1,25 @@
...
+  
+  cd gh-pages
+  git config --global user.email "travis@travis-ci.org"
+  git config --global user.name "travis-ci"
+  git rm -rf api/latest 
+  mv ${TARGET}/site/apidocs api/latest
+  git add -A -f api/latest
+  git commit -m "Lastest javadoc on successful travis build $TRAVIS_BUILD_NUMBER auto-pushed to gh-pages"
```
sameb: FYI I noticed this typo in my original guice script the other day -- s/Lastest/Latest

util/publish-snapshot-on-commit.sh
```diff
@@ -0,0 +1,12 @@
+# see https://coderwall.com/p/9b_lfq
+
+if [ "$TRAVIS_REPO_SLUG" == "google/truth" ] && \
```
sameb: wrong project name here too

util/generate-latest-docs.sh
cgruber: Ugh.  Yes.  Thanks sam.  You're totally correct.  I'll fix.  It sucks that you can't test this after-success stuff conditional on a particular branch like master until you actually merge it. :(

cgruber:
Rolling a fix for your observations.  This is what I get for doing stuff before coffee.

PR #204, opened 2015-03-17
Avoid side effects in downstream projects by including only relocated dependencies

DanielThomas:
Autovalue shades Apache libraries without relocation, meaning that they're not conflict resolved downstream and cause side effects in projects using autovalue. In this case, Velocity brings in commons-lang 2.4 causing https://issues.apache.org/jira/browse/LANG-421 in my project.

This change explicitly includes only the artifacts that are relocated to avoid these side effects. (I also upgraded guava-test so it matches the Guava version in use for good measure.)

### DIFF
--- a/value/pom.xml
+++ b/value/pom.xml
@@ -132,41 +132,41 @@
         </executions>
       </plugin>
       <plugin>
-	<groupId>org.apache.maven.plugins</groupId>
-	<artifactId>maven-shade-plugin</artifactId>
-	<version>2.3</version>
-	<executions>
-	  <execution>
-	    <phase>package</phase>
-	    <goals>
-	      <goal>shade</goal>
-	    </goals>
-	    <configuration>
-	      <minimizeJar>true</minimizeJar>
-          <artifactSet>
-            <!-- We don't include Apache classes because they often load classes from strings,
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>2.3</version>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <minimizeJar>true</minimizeJar>
+              <artifactSet>
+                <!-- We don't include Apache classes because they often load classes from strings,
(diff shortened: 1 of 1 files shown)


DanielThomas:
I've submitted the CLA.

value/pom.xml
```diff
@@ -139,33 +143,34 @@
...
 	      <minimizeJar>true</minimizeJar>
-              <filters>
-                <filter>
-                   <artifact>org.apache.velocity:*</artifact>
-                   <includes>
-                       <include>**</include>
-                   </includes>
-                </filter>
-              </filters>
+          <artifactSet>
```
eamonnmcmanus: Indentation looks wrong.

value/pom.xml
```diff
@@ -139,33 +143,34 @@
 	    </goals>
 	    <configuration>
 	      <minimizeJar>true</minimizeJar>
-              <filters>
-                <filter>
-                   <artifact>org.apache.velocity:*</artifact>
```
eamonnmcmanus: I'm confused. Does this mean that people will have to express an explicit dependency on this if they want to use AutoValue?

DanielThomas: It means that the filtered pom includes Velocity, so the Apache dependencies come in transitively, rather than being provided by the jar, so they participate in conflict resolution.

value/pom.xml
DanielThomas: The original indentation was off, and I fixed that, so that might be why it looks wonky?

eamonnmcmanus: Still looks off to me. `<artifactSet>` should be indented further than `<configuration>`.

DanielThomas: Oh right, there's definitely some other problems there - fixed now.

eamonnmcmanus:
I'm uncomfortable with the approach here. I feel that the AutoValue jar should be self-contained. Ideally we'd be able to shade everything in it, but Velocity doesn't play along with that. It turns out that Commons does, though, despite my previously concluding otherwise. So I'm proposing #205 as an alternative to this. That way people are only adversely affected if they are building with a different version of Velocity, as opposed to the current bad situation where they were also affected if they were building with a different version of Commons from the one Velocity was using.

PR #205, opened 2015-03-18
Change Maven shading for AutoValue so that org.apache.commons is shaded.

eamonnmcmanus:
I previously believed this was not possible due to classes being loaded by name, but it turns out that it is, provided we avoid tickling a bug in shading related to string constants.

### DIFF
--- a/value/pom.xml
+++ b/value/pom.xml
@@ -148,9 +148,18 @@
                 </filter>
               </filters>
 	      <relocations>
-		<!-- We don't relocate Apache classes because they often load classes from strings,
-                     and the relocator doesn't know to rewrite those strings. This applies to both
-                     Velocity and Commons. We also don't minimize Velocity for similar reasons. -->
+		<!-- We don't relocate Velocity classes because they often load classes from
+		     strings, and the relocator doesn't know to rewrite those strings.
+                     We also don't minimize it for similar reasons.
+		     You would expect that we could express this by relocating org.apache and
+		     excluding org.apache.velocity.**, but it turns out that doesn't work because
+		     the relocator for some reason feels the need to rewrite string constants that
+		     begin with org/apache even if they also begin with org/apache/velocity, and
+		     that throws resource name constants into the air. -->
+		<relocation>
+		  <pattern>org.apache.commons</pattern>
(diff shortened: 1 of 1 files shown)


value/pom.xml
```diff
@@ -148,9 +148,18 @@
...
-                     and the relocator doesn't know to rewrite those strings. This applies to both
-                     Velocity and Commons. We also don't minimize Velocity for similar reasons. -->
+		<!-- We don't relocate Velocity classes because they often load classes from
+		     strings, and the relocator doesn't know to rewrite those strings.
+                     We also don't minimize it for similar reasons.
+		     You would expect that we could express this by relocating org.apache and
+		     excluding org.apache.velocity.**, but it turns out that doesn't work because
+		     the relocator for some reason feels the need to rewrite string constants that
+		     begin with org/apache even if they also begin with org/apache/velocity, and
+		     that throws resource name constants into the air. -->
```
cgruber: Is there a bug number for tracking this exclusion failure? If so, maybe reference the link here, for future investigation.

value/pom.xml
```diff
@@ -148,9 +148,18 @@
                 </filter>
               </filters>
 	      <relocations>
-		<!-- We don't relocate Apache classes because they often load classes from strings,
-                     and the relocator doesn't know to rewrite those strings. This applies to both
-                     Velocity and Commons. We also don't minimize Velocity for similar reasons. -->
+		<!-- We don't relocate Velocity classes because they often load classes from
+		     strings, and the relocator doesn't know to rewrite those strings.
```
cgruber: God, this almost makes me want to go back to your original hand-rolled template. :)

cgruber:
LGTM

value/pom.xml
eamonnmcmanus: Wasn't able to find one. If I were more motivated I suppose I would log one.

value/pom.xml
eamonnmcmanus: I'm seriously tempted to write a template engine from scratch that is compatible with the subset of Velocity that I use. Velocity is great in that it's very widely used and has a great deal of tooling, but it's terribly overengineered for what we're using it for.

cgruber: Call it "NMPS" - nanometers per second. As in really small velocity.

PR #212, opened 2015-03-31
Adding support to AutoValue.Id.

rdtorres:
This new annotation should be used when an autoValue does not want to use all its fields in the equals() and hashcode() method. So, the autovalue bean can simple annotate all fields that should be used to unique identify that bean.

Let me know if you like it, so I can add more JUnit tests to cover edge cases.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/AutoValueIdsProcessor.java
+++ b/value/src/main/java/com/google/auto/value/processor/AutoValueIdsProcessor.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2014 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.auto.value.processor;
+
+import com.google.auto.common.MoreElements;
+import com.google.auto.common.SuperficialValidation;
+import com.google.auto.service.AutoService;
+import com.google.auto.value.AutoValue;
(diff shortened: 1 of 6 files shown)


rdtorres:
I used another email account to submit this pull request. How can I associate this other account to my CLA account ?

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -494,7 +504,7 @@ private void fixReservedIdentifiers(Map<ExecutableElement, String> methodToIdent
   }
 
   private String disambiguate(String name, Collection<String> existingNames) {
-    for (int i = 0; ; i++) {
```
rdtorres: This change has nothing to do with the commit. But I fixed it because checkstyle was complaining about it.

value/src/main/java/com/google/auto/value/processor/AutoValueIdsProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Annotation processor that checks that the type that {@link com.google.auto.value.AutoValue.Id}
+ * is applied to is nested inside an {@code @AutoValue} class. The actual code generation for ids
+ * is done in {@link AutoValueProcessor}.
+ *
+ * @author Rafael Torres
+ */
+@AutoService(Processor.class)
+public class AutoValueIdsProcessor extends AbstractProcessor {
```
rdtorres: I was not sure if I should create another processor to validate or if I should use AutoValueBuilderProcessor. My suggestion is to rename AutoValueBuilderProcessor to some generic name and use it to validate Builder, Validate and Id.

eamonnmcmanus:
We explicitly say in the [documentation](https://github.com/google/auto/tree/master/value#derived-fields) that we don't support fields that don't participate in equals and hashCode. If we did want to do so, this would certainly be one way of going about it. In that case I think we would also want the ability to say, independently, whether a field appears in toString.
Pinging @kevinb9n.

rdtorres:
I've read the documentation and I know that auto-value does not support it.
In fact,  this is the main reason behind this pull request: Add to support
to it =). The toString method can also follow a similar approach where an
annotation like "Autovalue.Printable" would indicate which fields should be
printed or not. Again, these annotations are extra, non-mandatory
annotations. So, default auto-model behaviour will be exactly the same as
we have nowadays...

Let me know what do you guys think.

Issue #82, opened 2014-02-24
Factory generation mixups when more than 1 factory in same package with implementing interface


Zubnix:
When 2 or more types with each their own autofactory and each autofactory with their own interface is declared, then each generated autofactory class contains the implementation of _all_ autofactory interfaces.

gjoseph:
Hey, I don't know if this is strictly related, but I'm noticing that if I have 2 classes, in different packages, both using the same `@AutoFactory(implementing = Foo.class)`, the second factory to be generated contains the interface's method twice. (and one of those has the signature of the _other_ type)

gjoseph:
Eh, in fact, both generated factories have the problem. So it goes like this: in my sources:
````
package foo
public interface Foo {}
public interface FooFactory {
Foo heyDoMe();
}

package foo.bar
@AutoFactory(implementing = FooFactory.class)
public class Bar implements Foo {}

package foo.qux
@AutoFactory(implementing = FooFactory.class)
public class Qux implements Foo {}
````

Both resulting factories like this, both have these 2 methods, which makes the compiler rather unhappy obviously.
````
[...]
  @Override
  public foo.bar.Bar heyDoMe() {
    return create(p);
  }
  @Override
  public Qux heyDoMe() {
    return create(p);
  }
````

busybeaver:
Sorry for being impatient... but do we already know when this bug will be fixed (and a new version of autofactory will be published)?
This bug makes it autofactory practically unusable... creating for each factory (which implements an interface) a separate package results in almost more time lost than saved due to the factory autogeneration.

PR #216, opened 2015-04-01
Fix issues with multiple factories that implement the same interface

gjoseph:
Assuming a bug in either auto-common or dagger-compiler. Since the latter depends on 1.0-SNAPSHOT of the former, it seems that having 0.3 in auto-factory messed things up. Set it back to snapshot here until a new release of auto-common is available and/or it is shaded into dagger-compiler.

Somehow I'd have expected this kind of problem to break the build, but this isn't the first time I notice an annotation processor that just silently does nothing.

### DIFF
--- a/factory/src/it/github-82/pom.xml
+++ b/factory/src/it/github-82/pom.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright (C) 2013 Google, Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


gjoseph:
Ok, so I actually wanted to do 2 different PRs, one for the build (since it meant I couldn't generate an ` auto-factory.jar`) and one for #82 but for some reason forgot to branch out. Let me know if you'd really rather have 2 individual ones ? 
(re: the fix-build commit, I wasn't actually able to really identify what was wrong, but it seemed to be somewhere in the depths of Dagger's `ComponentProcessor`)

gjoseph:
.. and the revert of 55bda7e shows the same results on Travis - target/classes is empty !

gjoseph:
Rebased PR on current master, which now depends on auto-common 0.4 (which should address the bug I had with the dependency to 0.3)

tcz:
What is this PR waiting for?

netdpb:
I believe this PR is obsolete at this point. If you still experience this issue, please file a new bug with a repro example.

Thanks!

Issue #116, opened 2014-07-08
Annotate constructor parameters with `@Nullable` where relevant


desseim:
This was already raised in the discussion of #73 , but what is the state of making the constructor arguments nullable for properties which getters are annotated with `@Nullable` ?
This makes total sense from a user standpoint IMO (e.g. for static analysis), and wouldn't present as many tricky decisions as `equals()` would: simply annotate the parameter with the annotation the getter was annotated with (and no new dependency pulled either).

sarod:
+1 here.

eamonnmcmanus:
This issue was fixed last year but we neglected to close it.

PR #220, opened 2015-04-04
Issue #116: Annotate constructor parameters with `@Nullable` where relevant

sarod:
Pull request for Issue #116: Annotate constructor parameters with `@Nullable` where relevant

If a "Nullable" annotation is found on a property it is now also used for the constructor parameters.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/autovalue.vm
+++ b/value/src/main/java/com/google/auto/value/processor/autovalue.vm
@@ -35,7 +35,7 @@ final class $subclass$formalTypes extends $origClass$actualTypes {
   $subclass(
 #foreach ($p in $props)
 
-      $p.type $p #if ($foreach.hasNext) , #end
+     #if ($p.nullable) ${p.nullableAnnotation} #end $p.type $p #if ($foreach.hasNext) , #end
 #end ) {
 #foreach ($p in $props)
   #if (!$p.kind.primitive && !$p.nullable)
--- a/value/src/test/java/com/google/auto/value/processor/PropertyAnnotationsTest.java
+++ b/value/src/test/java/com/google/auto/value/processor/PropertyAnnotationsTest.java
@@ -99,7 +99,8 @@ public class PropertyAnnotationsTest extends TestCase {
     return JavaFileObjects.forSourceLines("foo.bar.Baz", lines);
   }
 
-  private JavaFileObject expectedCode(List<String> annotations) {
+  private JavaFileObject expectedCode(List<String> annotations, String expectedConstructorParamAnnotation) {
+    String constructorParamPrefix = expectedConstructorParamAnnotation != null ? expectedConstructorParamAnnotation + " " : "";
     ImmutableList<String> list = ImmutableList.<String>builder()
         .add(
             "package foo.bar;",
@@ -110,7 +111,7 @@ public class PropertyAnnotationsTest extends TestCase {
             "final class AutoValue_Baz extends Baz {",
             "  private final int buh;",
             "",
-            "  AutoValue_Baz(int buh) {",
+            "  AutoValue_Baz("+constructorParamPrefix+"int buh) {",
             "    this.buh = buh;",
             "  }",
             ""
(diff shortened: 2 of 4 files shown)


sarod:
I Signed it!

eamonnmcmanus:
Thanks! This looks right to me, and I only have minor style comments.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
```diff
@@ -35,7 +35,7 @@ final class $subclass$formalTypes extends $origClass$actualTypes {
   $subclass(
 #foreach ($p in $props)
 
-      $p.type $p #if ($foreach.hasNext) , #end
+     #if ($p.nullable) ${p.nullableAnnotation} #end $p.type $p #if ($foreach.hasNext) , #end
```
eamonnmcmanus: Can you check, if you haven't already, that the indentation is correct for both `@Nullable` and plain parameters? Each line should be indented four spaces further than the start of the constructor declaration.

(The need to tweak templates minutely like this is the main reason I might end up abandoning the template approach, much as I like it otherwise.)

value/src/test/java/com/google/auto/value/processor/PropertyAnnotationsTest.java
```diff
@@ -99,7 +99,8 @@ private JavaFileObject sourceCode(List<String> imports, List<String> annotations
     return JavaFileObjects.forSourceLines("foo.bar.Baz", lines);
   }
 
-  private JavaFileObject expectedCode(List<String> annotations) {
+  private JavaFileObject expectedCode(List<String> annotations, String expectedConstructorParamAnnotation) {
+    String constructorParamPrefix = expectedConstructorParamAnnotation != null ? expectedConstructorParamAnnotation + " " : "";
```
eamonnmcmanus: Google style imposes a 100-column limit. Can you split this line? There might be a long line in CompilationTest too.

value/src/test/java/com/google/auto/value/processor/PropertyAnnotationsTest.java
```diff
@@ -110,7 +111,7 @@ private JavaFileObject expectedCode(List<String> annotations) {
             "final class AutoValue_Baz extends Baz {",
             "  private final int buh;",
             "",
-            "  AutoValue_Baz(int buh) {",
+            "  AutoValue_Baz("+constructorParamPrefix+"int buh) {",
```
eamonnmcmanus: Please put spaces around each `+`.

value/src/test/java/com/google/auto/value/processor/PropertyAnnotationsTest.java
```diff
@@ -152,26 +153,37 @@ private JavaFileObject expectedCode(List<String> annotations) {
     return JavaFileObjects.forSourceLines("foo.bar.AutoValue_Baz", lines);
   }
 
+  
   private void assertGeneratedMatches(
       List<String> imports,
       List<String> annotations,
-      List<String> expectedAnnotations) {
+      List<String> expectedAnnotations, String expectedConstructorParamAnnotation) {
```
eamonnmcmanus: Please put the new parameter on a line by itself.

value/src/test/java/com/google/auto/value/processor/PropertyAnnotationsTest.java
```diff
@@ -152,26 +153,37 @@ private JavaFileObject expectedCode(List<String> annotations) {
...
   }
 
   public void testSimpleAnnotation() {
     assertGeneratedMatches(
         ImmutableList.of("import javax.annotation.Nullable;"),
         ImmutableList.of("@Nullable"),
-        ImmutableList.of("@javax.annotation.Nullable"));
+        ImmutableList.of("@javax.annotation.Nullable"),
+        "@javax.annotation.Nullable"
+        );
```
eamonnmcmanus: Please join this to the end of the previous line. (I find that the assertGeneratedMatches method is now quite cryptic but that's a matter for another day.)

value/src/main/java/com/google/auto/value/processor/autovalue.vm
sarod: It seems ok. 
I put a break point in AutoValueProcessor.processType to verify manually while executing the unit tests. Let me know if this is not the right way to verify this or if special test cases should be added.

sarod:
Thanks for the quick feedback.
I pushed changes to match your comments on style. Let me know if something is missing.  

Please note that I tried to run checkstyle using `mvn checkstyle:checkstyle` but when running it at the root of auto it doesn't seem to do much and when running it from "value" directory it complains that there is no "checkstyle.xml" file.

eamonnmcmanus:
Excellent, thanks!

Issue #81, opened 2014-02-23
Provide OSGi meta data


rherrmann:
To ease using e.g. AutoValue in an OSGi environment it would help if the released jar had a MANIFEST.MF containing OSGi meta data. The library/bundle could then be provisioned in a p2 repository.

The [Bundle Plugin for Maven](http://felix.apache.org/site/apache-felix-maven-bundle-plugin-bnd.html) is one option to generate the meta data. Given that the number and names of public packages of the auto libraries don't change often, the maintenance effort should be very low.

sarod:
It would be nice t have this fixed for the release of auto-value 1.1 (#224)

agarwalnitish18:
Gitcolony notification failed - java.lang.IllegalArgumentException: Invalid url: 

how to fix this error??

cgruber:
@sarod  - Can you test this in the snapshot? It's using auto service 1.0-rc2, and similar windows issues in dagger were resolved by this. 

@agarwalnitish18 - what on earth are you talking about?

sarod:
@cgruber I think there was some confusion here. I was referring to the initial problem for which this issue was created (the lack of OSGI metadata) in both my comment and my pull request. I was NOT referring to the bug related to path on windows. Sorry if this wasn't clear, it may be better to split this github issue because it covers 2 unrelated things.

Anyway, unfortunately I will not be able to test the path related bug on windows because I'm using ubuntu.

Is there any chance to see my pull request about osgi metadata merged before auto-value 1.1 is released?

cgruber:
Apparently it's not possible, since we pushed 1.2.  Sorry, I've not been involved in the content of AutoValue, just the technical work of releases and syncing, etc.  We should triage this and get it addressed.

PR #228, opened 2015-04-12
Fix AutoFactory build

tbroyer:

PR #229, opened 2015-04-13
Pull request for Issue #81Provide OSGi meta data

sarod:
Pull request for Issue #81.

Provide OSGI manifest attributes for auto-value, auto-factory and auto-service using maven-bundle-plugin.

Note that auto-factory does not compile currently. It complains about missing dagger  2.0-SNAPSHOT dependency I tried using the 2.0 but it then fails with compile errors. So I was not able to test the proposed changes are not tested for auto-factory.

### DIFF
--- a/factory/pom.xml
+++ b/factory/pom.xml
@@ -90,6 +90,36 @@
   <build>
     <plugins>
       <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <version>2.5</version>
+        <configuration>
+          <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+          </archive>   
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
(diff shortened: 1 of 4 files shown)


sarod:
Travis Build fails because dagger 2.0-SNAPSHOT is missing. 
(Pull Request #228 is supposed to fix this)

cgruber:
AutoFactory has now removed dependency on dagger, so the build should work just fine.

sarod:
Is there any chance of this being merged before release of auto-value 1.1?

sarod:
This pull request is probably outdated by now. 
@cgruber I didn't get any feedback on the original pull request should I rebase it ? Are you accepting contributions on this issue?

eamonnmcmanus:
What do you see as the principal benefits of this change?

sarod:
@eamonnmcmanus sorry for the late replay I missed your answer.

OSGI manifest attributes contain runtime metadata. So given the fact that there is no reference to auto-value at runtime it could seem useless to include OSGI metadata.

However some OSGI development environment like eclipse (PDE) uses the OSGI metadata to build the classpath used for compilation. 
So the lack of OSGI metadata on autovalue makes auto-value unusable in PDE.

Issue #202, opened 2015-03-09
Extensibility API for AutoValue


eamonnmcmanus:
Issues #87, #162, and #201 are examples of use cases that are a bit too specialized to put into the core of AutoValue, but which are still important when they apply. We should define an extensibility API so that users can build extensions that run at the same time as AutoValue. The general idea would be that there would be an interface, say AutoValueExtension, and that the AutoValue processor would use [ServiceLoader](http://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) to find implementations of this interface. So if you include on your `-processorpath` a jar that contains such an implementation, AutoValue will invoke it when it runs.

A possible sketch of the interface is this:

```java
public interface AutoValueExtension {
  interface Context {
    ProcessingEnvironment processingEnvironment();
    String packageName();
    TypeElement autoValueClass();
    Map<String, ExecutableElement> properties();
  }

  boolean applicable(Context context);

  String generateClass(Context context, String classToExtend, String classToImplement);
}
```

The idea is that every extension gets its `applicable` method called, and those that return true get their `generateClass` method called. That method can return a String that is an implementation of the class called classToImplement that extends the class called classToExtend; or it can return null (for example if it has generated some related code but doesn't need to intervene in the AutoValue hierarchy). So if you have two extensions, `@AutoValue class Foo` might result in a hierarchy `AutoValue_Foo extends $AutoValue_Foo extends $$AutoValue_Foo extends Foo`, where `AutoValue_Foo` is generated by one of the extensions, `$AutoValue_Foo` is generated by the other, and `$$AutoValue_Foo` contains the logic generated by the AutoValue processor. This means that extensions can override the implementations of `toString()` etc.

PhilippWendler:
As far as I understand, this would allow me to customize the object creation, too, because I can insert code into the constructor of the class that my extension generates, correct? This would be good.

The use case I have is that I want to automatically create an `@AutoValue` instance from something that wraps a `Map<String, String>`. We use something like this for configuration options (currently based on reflection instead of AutoValue).

So my constructor would not have parameters for all properties but only for the map.
In the constructor, I could retrieve the value for each property from the map, convert it to the expected type, and pass them to the super constructor. The down side would be that yet another class gets created that is actually not necessary if I could intercept the passing of values from the factory method to the AutoValue-generated constructor somehow without it (because my subclass would not add or override any methods, only the constructor).

rharter:
[...]

PR #237, opened 2015-05-15
Extensibility

rharter:
This is a first pass implementation of issue #202, providing two options for extending AutoValue through the use of compile time plugins.

The first method, is by having your extension create an intermediate class between the abstract template class and the final AutoValue implementation of the class.   The provides a good mechanism for adding functionality to an AutoValue class, while still keeping the implementation segregated.

The second method is to alter the final AutoValue class itself.  This was required to support Android Parcelable implementations, as Parcels are required to have static factory objects, and this isn't achievable via inheritance.  Using this method, extensions have to option of adding imports, interfaces and code to the resulting AutoValue class implementation.  You can see a working example of one such plugin [here](https://github.com/rharter/auto-parcel).

You can view a test Android app that makes use of the auto-parcel extension [here](https://github.com/rharter/autoexttest).


rharter:
I signed it!

eamonnmcmanus:
I think the Travis failure indicates that you need to merge in recent changes.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.lang.model.element.TypeElement;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Created by rharter on 5/1/15.
+ */
+public interface AutoValueExtension {
+
+  public interface Context {
```
eamonnmcmanus: The public modifier is redundant.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,33 @@
...
+    ProcessingEnvironment processingEnvironment();
+    String packageName();
+    TypeElement autoValueClass();
+    Map<String, ExecutableElement> properties();
+  }
+
+  public interface GeneratedClass {
+    String className();
+    String source();
+    Collection<String> additionalImports();
```
eamonnmcmanus: I think it would make sense for this to be something like `Set<TypeElement>` referencedTypes(). Also, the methods in this interface that return `Collection<Foo>` should instead return `Set<Foo>` since semantically I think that's what's required.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,33 @@
...
+    Collection<ExecutableElement> consumedProperties();
+    Collection<String> additionalInterfaces();
+    Collection<String> additionalCode();
```
eamonnmcmanus: Is this a list of lines?

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Created by rharter on 5/1/15.
+ */
+public interface AutoValueExtension {
```
eamonnmcmanus: The overall structure looks OK but it needs javadoc.

eamonnmcmanus: Actually you might want to hold off on the javadoc. The fact that the proposed model doesn't fit the Parcelable use case all that well makes me wonder whether we should stay with it.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
rharter: No, this is a collection of blocks of code.  This is a collection because the processor collects all of the additional code from all extensions to be added to the class, and I made it a collection here thinking that an extension implementor might have different possible chunks of code that need to be added based on the AutoValue class specification.  This could certainly be changed to simply a string, if that's more straightforward.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
rharter: This was implemented as strings to alleviate the need for extra dependencies.  For instance, in my Parcelable example, Parcelable and Parcel are both classes in the Android Framework.  By adding them as strings here, my plugin isn't dependent on the Android SDK, as these String FQCNs can simply be added to the imports clause.  AFAIK I can't create a TypeElement without having access to the represented class.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
eamonnmcmanus: I'm actually not really understanding how this applies in the Parcelable case. Say you have no other extensions. Then the Parcelable extension will generate AutoValue_Foo which extends $AutoValue_Foo which extends Foo. $AutoValue_Foo is the AutoValue implementation class. Isn't it AutoValue_Foo that needs to have the definition of CREATOR and so on?

value/src/main/java/com/google/auto/value/AutoValueExtension.java
eamonnmcmanus: [Elements.getTypeElement(CharSequence)](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/util/Elements.html#getTypeElement-java.lang.CharSequence-)

rharter: Great, I'll update it.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
rharter: When I implemented the subclassing version of this, the ordering was actually different.  I treated the extension classes as the intermediaries, so AutoValue_Foo is the AutoValue implementation, which extends $AutoValue_Foo which is the AutoParcel implementation, which extends Foo.

This is important, particularly if you have more than one extension that requires things like static factory methods or static properties.  Even if you have multiple plugins and only one requires static properties, we'd need to devise a prioritization mechanism for the extensions.

That's why I implemented this with both the ability to add intermediate classes, and the ability to add code to the final concrete AutoValue class.

eamonnmcmanus: It's important for extensions to be subclasses of the generated AutoValue implementation, so they can override the generated toString() etc. One of our use cases is exactly that (being able to omit certain properties from toString()).

There's nothing stopping extensions from having static properties or methods even if they are not at the bottom of the inheritance hierarchy. The issue with Parcelable seems to be that the CREATOR field has to be in the Parcelable class itself and can't be inherited from an ancestor. I personally think that design is wrong, but there it is. We're certainly [not the first](http://stackoverflow.com/questions/12809928/parcelable-and-inheritance) to encounter it.

So in conclusion I think we probably want a way for an extension to declare that it must be at the bottom of the hierarchy? If you have two such extensions and they both want to generate code for a given class then we'll just have to reject that.

The alternative design that we considered was to switch the AutoValue implementation so it uses [JavaPoet](https://github.com/square/javapoet) instead of templates, and have extensions be able to contribute code to the single generated class by calling JavaPoet methods. I've been reluctant to do that for a number of reasons. First, I like templates better than code-generating-code. Second, it doesn't give a clean solution for the toString() override. Third, it feels kind of anarchic when there is more than one extension contributing code (especially if they can change the already-contributed code). Finally, it risks having extensions know too much about exactly what the AutoValue implementation class looks like.

Despite all that, we may conclude from this exercise that the JavaPoet approach is the way to go.

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -585,13 +674,6 @@ private static void determineObjectMethodsToGenerate(
             errors = true;
           }
           toImplement.add(method);
-        } else {
-          // This could reasonably be an error, were it not for an Eclipse bug in
-          // ElementUtils.override that sometimes fails to recognize that one method overrides
-          // another, and therefore leaves us with both an abstract method and the subclass method
-          // that overrides it. This shows up in AutoValueTest.LukesBase for example.
-          errorReporter.reportWarning("@AutoValue classes cannot have abstract methods other than"
```
eamonnmcmanus: Is there logic to make this check once extensions have had a chance to claim abstract methods? I didn't see it.

rharter: Not at this point. This will probably have to be added back in.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
rharter: It sounds like the JavaPoet approach is fairly close to what I had added for the Parcelable support, basically allowing extensions to add code directly to the AutoValue class.  I can definitely appreciate the apprehension about that, I was also initially apprehensive, but my inverted inheritance didn't really offer a solution and it didn't seem like ordering extensions was a good solution. 

Also, I completely agree that the CREATOR static field is an unfortunate approach.  I'd love to see something like an annotation on the class pointing to a factory class.  That's kind of what this extension would provide, only generating the factory.  Unfortunately, other [solutions](https://github.com/johncarl81/parceler) require wrapping, which I'm hoping this can avoid.

Regarding the inverted inheritance (AutoValue_Foo (from AutoValue) inherits from $AutoValue_Foo (from extension) inherits from abstract Foo), I don't think that causes problems for the toString use case, provided the AutoValue processor waits until after the extensions are processed to determine methods (toString, equals, hashCode) to override.  Then, if there is an extension that provides a custom implementation of toString() it would be the same as if the user wrote a custom toString() in the abstract template, causing AutoValue not to generate it's own implementation, right?

eamonnmcmanus: Well yes, but that would mean AutoValueProcessor would need to parse the code output by extensions to see whether it defines toString(). That can be done (by deferring the generation of the final AutoValue implementation to the next round of annotation processing) but it's not simple. I'm also not keen on a solution that is a mix of generating one class per extension and generating further code to be injected into the original implementation class.

Incidentally another issue with the JavaPoet approach is that extensions would need to apply the same Maven-shading as AutoValue so that their references to JavaPoet APIs would be match the APIs that AutoValue would be using. In other words, if AutoValue renames com.squareup.javapoet.TypeName to autovalue.com.squareup.javapoet.TypeName, then extensions will either have to reference that renamed class in their source code, or apply the same renaming through Maven shading when packaged.

rharter: Yeah, I don't like the JavaPoet approach, it also adds an extra layer of requirements around implementation details on the extension developer.

I think this discussion is good, and it seems to point to your original inheritance model, with a mechanism to require an extension be at the end of the chain, and a limit that only one such extension can be present at a time.

Do you think it makes sense for me to note these finding in the issue, cancel this PR, and alter the implementation to remove the additional code and force extensions to create subclasses of the AutoValue generated class, then submit a new PR?

eamonnmcmanus: Yes, that might be the best approach.

kevinb9n: Ryan, I just wanted to chime in and say how much Eamonn and I appreciate
the work you're doing and the forward momentum you're providing for this
important feature!



On Mon, May 18, 2015 at 12:43 PM, Ryan Harter <notifications@github.com>
wrote:

> In value/src/main/java/com/google/auto/value/AutoValueExtension.java
> <https://github.com/google/auto/pull/237#discussion_r30540505>:
>
> > +public interface AutoValueExtension {
> > +
> > +  public interface Context {
> > +    ProcessingEnvironment processingEnvironment();
> > +    String packageName();
> > +    TypeElement autoValueClass();
> > +    Map<String, ExecutableElement> properties();
> > +  }
> > +
> > +  public interface GeneratedClass {
> > +    String className();
> > +    String source();
> > +    Collection<String> additionalImports();
> > +    Collection<ExecutableElement> consumedProperties();
> > +    Collection<String> additionalInterfaces();
> > +    Collection<String> additionalCode();
>
> Yeah, I don't like the JavaPoet approach, it also adds an extra layer of
> requirements around implementation details on the extension developer.
>
> I think this discussion is good, and it seems to point to your original
> inheritance model, with a mechanism to require an extension be at the end
> of the chain, and a limit that only one such extension can be present at a
> time.
>
> Do you think it makes sense for me to note these finding in the issue,
> cancel this PR, and alter the implementation to remove the additional code
> and force extensions to create subclasses of the AutoValue generated class,
> then submit a new PR?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/auto/pull/237/files#r30540505>.
>



-- 
Kevin Bourrillion | Java Librarian | Google, Inc. | kevinb@google.com

rharter:
@eamonnmcmanus I think this is in a good state to check out.  

I'm still having a problem with my Parcelable extension in that the annotated class has to implement the Parcelable interface, but AutoValue takes the `int describeContents()` method from the interface to be something that it should generate.  I understand the need to include inherited abstract methods in the properties list, but I'm not sure how to tell AutoValue that it doesn't need to worry about this particular interface method with no parameters that returns an int.  For now, I'm working around this by simply implementing the super simple version in my annotated class.  The generated class also generates an implementation.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * An AutoValueExtension allows for extra functionality to be created during the generation
+ * of an AutoValue class.
+ *
+ * Extensions are discovered at compile time using the {@link java.util.ServiceLoader} APIs,
```
eamonnmcmanus: Can you insert a `<p>` at the start of the first line of each paragraph here (except the first)? Like this:
```java
 * <p>Extensions are discovered...
```

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * The context of the generation cycle.
+   */
+  public interface Context {
+
+    /**
+     * The processing environment of this generation cycle.
+     *
+     * @return The {@link javax.annotation.processing.ProcessingEnvironment} of this generation cycle.
```
eamonnmcmanus: I think just `{@link ProcessingEnvironment}` would be fine, or just omit `{@link}` since it is the return type of the method.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * The package name of the classes to be generated.
+     *
+     * @return The package name of the classes to be generated.
+     */
+    String packageName();
+
+    /**
+     * The annotated class that this generation cycle is based on.
```
eamonnmcmanus: Maybe add a sentence to make it crystal clear, like this:
```
 * Given {@code @AutoValue public class Foo {...}}, this will be {@code Foo}.
```

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @return The ordered collection of properties.
+     */
+    Map<String, ExecutableElement> properties();
+  }
+
+  /**
+   * Determines whether this extension applies to the given context.
+   *
+   * @param context The {@link com.google.auto.value.AutoValueExtension.Context} of the code
```
eamonnmcmanus: Again I think `{@link Context}` or no `{@link}` would be fine.

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -17,36 +17,23 @@
...
+import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
-import com.google.common.collect.FluentIterable;
-import com.google.common.collect.ImmutableBiMap;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
+import com.google.common.collect.*;
```
eamonnmcmanus: Maybe your IDE did this, but please leave the explicit imports.

eamonnmcmanus:
I'm still working through this. More tomorrow.

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -94,9 +87,14 @@ public SourceVersion getSupportedSourceVersion() {
    */
   private final List<String> deferredTypeNames = new ArrayList<String>();
 
+  private Iterable<AutoValueExtension> extensions;
+
   @Override
   public synchronized void init(ProcessingEnvironment processingEnv) {
     super.init(processingEnv);
+    if (extensions == null) {
+      extensions = ServiceLoader.load(AutoValueExtension.class, getClass().getClassLoader());
```
eamonnmcmanus: I think I would do this directly in the no-arg constructor, i.e.
```java
this(ServiceLoader.load(AutoValueExtension.class, getClass().getClassLoader()));
```

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -382,26 +380,82 @@ private void processType(TypeElement type) {
...
+        return;
+      }
+      source = Reformatter.fixup(source);
+      writeSourceFile(fqClassName, source, type);
+
+      extClass = className;
+    }
+  }
+
+  private AutoValueExtension.Context makeExtensionContext(final TypeElement type, final Iterable<ExecutableElement> methods) {
```
eamonnmcmanus: We try to respect a 100-column limit.

value/src/main/java/com/google/auto/value/processor/AutoValueTemplateVars.java
```diff
@@ -84,6 +84,10 @@
   /** The simple name of the generated subclass. */
   String subclass;
 
+  /** True if the generated class should be final (there are no extensions that
+   * will generate subclasses) */
```
eamonnmcmanus: Nit: can you put `*/` on the next line?

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
```diff
@@ -0,0 +1,265 @@
...
+
+      boolean first = true;
+      for (Map.Entry<String, ExecutableElement> el : context.properties().entrySet()) {
+        if (first) {
+          first = false;
+        } else {
+          constructor.append(", ");
+        }
+
+        // TODO How are we going to handle the constructor?
```
eamonnmcmanus: What is the problem?

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
```diff
@@ -0,0 +1,265 @@
...
+        "foo.bar.Baz",
+        "package foo.bar;",
+        "",
+        "import com.google.auto.value.AutoValue;",
+        "",
+        "@AutoValue",
+        "public abstract class Baz {",
+        "  abstract String foo();",
+        "}");
+    JavaFileObject expectedOutput = JavaFileObjects.forSourceLines(
```
eamonnmcmanus: I think I'd prefer not to have any new expected-output strings for the generated AutoValue implementation. They are a pain to maintain during changes. I think it would be enough just to assert on expectedExtensionOutput.

rharter: I understand what you're saying, but the intention of testing that output is to ensure that the main AutoValue output doesn't include the `final` qualifier and has the appropriate `$` prefix when extensions are present.  Do you think that's skippable?

eamonnmcmanus: Yes. If it doesn't get that right then the generated code won't compile and the test will fail.

rharter:
@eamonnmcmanus Is there anything else that needs to be done with this?

eamonnmcmanus:
There's a bunch of comments from July 8th that you haven't addressed, I think?

rharter:
Sorry, I guess I didn't comment when those were resolved, but they should all be resolved in the commits from that same day.

eamonnmcmanus:
OK, I'll take a look later today. Sorry about the misunderstanding.

value/com/example/annotations/AutoAnnotation_Erroneous_newEmpty.java
```diff
@@ -0,0 +1,40 @@
+
```
eamonnmcmanus: What is this?

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * An AutoValueExtension allows for extra functionality to be created during the generation
+ * of an AutoValue class.
+ *
+ * <p>Extensions are discovered at compile time using the {@link java.util.ServiceLoader} APIs,
+ * allowing them to run without any additional annotations.
+ *
+ * <p>Extensions can extend the AutoValue implementation by generating subclasses of the AutoValue
+ * generated class. It's not guaranteed that an Extension's generated class will be the final
+ * class in the inheritance hierarchy, unless it's {@link #mustBeAtEnd} method returns true,
```
eamonnmcmanus: Typo: its

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -0,0 +1,101 @@
...
+
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * An AutoValueExtension allows for extra functionality to be created during the generation
+ * of an AutoValue class.
```
eamonnmcmanus: Can you add a sentence:
NOTE: the design of this interface is not final and subject to change.

eamonnmcmanus:
I'm expecting @cgruber to make a PR shortly that will export Google's internal changes. Merging with those changes won't be too hard, I hope, and after that I think we should be good to go.

rharter:
Changes are in.  I think that erroneous empty file is the result of a test and was added through a careless `git add .`.  Seems wrong that tests would create stuff outside the target dir.

eamonnmcmanus:
Regarding the erroneous file, it must have been created by AutoAnnotationCompilationTest. That creates the file using File.createTempFile — is it possible that your system creates temporary files in the current directory??

rharter:
That's certainly possible.  My system is a Mac, I haven't tested the default File.createTempFile functionality there.

eamonnmcmanus:
Meanwhile we're running into issues merging in the latest Google changes. Those changes are unfortunately more extensive than I thought, so you might be in for a bit of merging work when they go in. I'm reluctant to merge this PR first, though, for fear of making that other merge even worse.

eamonnmcmanus:
Could you resolve the merge conflicts? Then I'll have another look. Hopefully Travis issue https://github.com/travis-ci/travis-ci/issues/4621 will be solved soon so we'll be able to get a correct indication of test status.

rharter:
Merged and passing tests locally.

rharter:
@eamonnmcmanus is the travis build hung?  Says it's still building.

eamonnmcmanus:
Seems Travis has been very slow of late. I waited about as long as that yesterday for a build.

rharter:
Looks like Travis is back up and all checks pass.

eamonnmcmanus:
Thanks for your patience with this!

frankiesardo:
Thank you both this CR is well worth the wait!

eamonnmcmanus:
@frankiesardo are you able to obtain the same results with this as with your `@AutoParcelable` fork?

frankiesardo:
I pushed a branch that uses the new Extension api. So far so good.

The only problem is related to the Parcelable interface method `int describeComponents()` as @rharter already pointed out here https://github.com/rharter/auto-parcel/commit/bfaba143737bd820d4b6c045f93446456591c08c#commitcomment-12614380).

Forcing the user to implement it, even if it's a simple method, it's an annoyance. Not sure if AutoValue can do something about it.

frankiesardo:
In #202 you mention briefly the possibility for an extension to claim a list of abstract methods. That would be helpful in this case.

rharter:
I'm working on that.

PR #238, opened 2015-05-15
Make generated classes have package-private constructors

### LINKED ISSUE #236: AutoFactory constructors are public if the class is public
It rarely makes sense to make the @Inject constructor of an AutoFactory public, but it's the default.

charlesmunger:
Fixes #236.


charlesmunger:
I'm a google employee, I don't need to sign the CLA.

cgruber:
It's the bot, Charles.  It doesn't (yet) handle google employees.

cpovirk:
I see that https://github.com/google/auto/issues/236 is closed, so I'll close this, too. I don't know that we reached a particularly authoritative conclusion there, but let's funnel the conversation to one place.

PR #244, opened 2015-05-24
AutoService: Add support for multiple service in one service provider

TomLottermann:
In my project I have many very small services which I Implemented in one class. This is paticularly useful if these services share a common interface.

Sadly however I found out that AutoService does not support that case.

In this commit I added the possiblity to specify multiple services in a single `@AutoService` annotation.


LookaPW:
Will this ever be merged? I am looking forward to this pull request.

willdurand:
This improvement would be awesome! 👍

henryptung:
Also looking forward to this.

eamonnmcmanus:
This functionality was implemented by #548.

PR #246, opened 2015-06-01
Add support for builder inheritance

redowl:
## Problem

Though inheritance is (understandably) discouraged, it can make sense in some cases, and is supported by the `@AutoValue` root classes.  However, it is not so easily supported by the `@AutoValue.Builder` classes.

Because of a type-equivalence check during pre-processing of the setter methods in an auto-value builder, any inherited methods on a builder are not able to return the appropriate type, regardless of use of generics, etc.

## Use Case

For a contrived use case, say that I want to model different types of communications in my application, all of which share a set of common properties, but all of which specialize.  I might create a set of model classes (using inheritance) like:

### Communciation.java
```java
package com.test.model;

import org.joda.time.DateTime;
import java.util.List;

interface Communication {
    public String type();
    public DateTime timestamp();
    public String sender();
    public List<String> recipients();
    public String content();

    static interface Builder<T extends Builder<T>> {
        public T type(String type);
        public T timestamp(DateTime timestamp);
        public T sender(String sender);
        public T recipients(List<String> recipients);
        public T content(String content);
    }
}
```

### Email.java
```java
package com.test.model;

import com.google.auto.value.AutoValue;
import java.util.List;

@AutoValue
public abstract class Email implements Communication {
     /* Package-private to prevent an undocumented public constructor */
    Email(){}

    public List<String> cc();
    public List<String> bcc();
    public String header();

    @AutoValue.Builder
    public abstract static class Builder implements Event.Builder<Builder> {
        /* Package-private to prevent an undocumented public constructor */
        Builder(){}

        public Builder cc(List<String> cc);
        public Builder bcc(List<String> bcc);
        public Builder header(String header);
    }
}
```

### DIFF
--- a/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueTest.java
+++ b/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueTest.java
@@ -1595,4 +1595,41 @@ public class AutoValueTest extends TestCase {
         methodInSuperclass.getAnnotation(HairyAnnotation.class);
     assertEquals(annotationInSuperclass, annotationInSubclass);
   }
+
+  public interface BuilderInterface{
+    public ImmutableList<String> baseStrings();
+
+    public interface Builder<T extends Builder<T>> {
+      T baseStrings(ImmutableList<String> baseStrings);
+    }
+  }
+
+  @AutoValue
+  public static abstract class BuilderWithInterface implements BuilderInterface{
+    public abstract ImmutableList<String> childStrings();
+
+    public static Builder builder() {
+      return new AutoValue_AutoValueTest_BuilderWithInterface.Builder();
+    }
+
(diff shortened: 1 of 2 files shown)


redowl:
I signed it!  It's signed as a corporation: RedOwl Analytics

eamonnmcmanus:
We're aware of the issue and tracking it internally. I'm not sure that the proposed fix is adequate, though. It should work for the subclass not to have to restate the inherited methods, but making that check properly is relatively hard, because any type variables on the builder type are distinct from the type variables of the final `@AutoValue` class, and also because `Types.asMemberOf` historically didn't work correctly in the Eclipse compiler.

eamonnmcmanus:
The underlying issue should be fixed in the latest snapshot.

PR #252, opened 2015-06-15
hashcode of Object property h fails

cranberyxl:
`h` is effectively a reserved property if it's an Object because it's used in the generated hashCode method

`error: int cannot be dereferenced`


eamonnmcmanus:
Excellent point about `h`. However, I think the best solution is to reference properties in `hashCode()` as `this.p` etc.

cranberyxl:
Would you rather see `this.property` for all properties and change all of the affected tests or just add `this.h` for the property `h`?

eamonnmcmanus:
I think all properties should be referenced as `this.property`. Alternatively, just call the local variable `h$` rather than `h`. Also, rather than adding a new test case to `CompilationTest`, it would be enough to show in `AutoValueTest` that an `@AutoValue` class with a property `h()` does compile.

cranberyxl:
Updated tests and approach

value/src/test/java/com/google/auto/value/processor/CompilationTest.java
```diff
@@ -1678,4 +1678,145 @@ public void testReferencingGeneratedClass() {
         .processedWith(new AutoValueProcessor(), new FooProcessor())
         .compilesWithoutError();
   }
+
+  public void testHashCode() {
```
eamonnmcmanus: I'd rather not have this extra test method. It is that much more code that we have to update every time we change the code template. The new AutoValueTest test case should be enough.

value/src/it/functional/src/test/java/com/google/auto/value/AutoValueTest.java
```diff
@@ -1595,4 +1595,16 @@ public void testCopyAnnotations() throws Exception {
...
+
+  @AutoValue
+  abstract static class HProperty {
+    public abstract Object h();
+    public static HProperty create(Object h) {
+      return new AutoValue_AutoValueTest_HProperty(h);
+    }
+  }
+  public void testHProperty() throws Exception {
+    // h is the hashCode variable, if we have a property named h, we get "error: int cannot be dereferenced"
```
eamonnmcmanus: Can you reword this comment so that it makes sense after the change? Something like:
```java
  // Checks that we can have a property called `h`. The generated hashCode() method has
  // a local variable of that name so there could be clashes if we aren't careful.
```

cranberyxl:
Test and comment changed

eamonnmcmanus:
Thanks!

PR #257, opened 2015-07-17
Add GWT integration tests and enhance GWT support

tbroyer:


value/src/it/gwtserializer/src/test/java/com/google/auto/value/client/GwtSerializerTest.java
```diff
@@ -0,0 +1,263 @@
...
+import com.google.common.annotations.GwtCompatible;
+import com.google.common.annotations.GwtIncompatible;
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.junit.client.GWTTestCase;
+import com.google.gwt.user.client.rpc.AsyncCallback;
+import com.google.gwt.user.client.rpc.RemoteService;
+import com.google.gwt.user.client.rpc.RemoteServiceRelativePath;
+import com.google.gwt.user.server.rpc.RemoteServiceServlet;
+
+public class GwtSerializerTest extends GWTTestCase {
```
eamonnmcmanus: This is great! We do actually have a GWT serialization test already but it is not public because it relies on some test infrastructure that is not published. Anyway it is much less thorough than this.

value/src/main/java/com/google/auto/value/processor/gwtserializer.vm
```diff
@@ -17,14 +17,13 @@ package $pkg;
...
+import com.google.gwt.user.client.rpc.CustomFieldSerializer;
 import com.google.gwt.user.client.rpc.SerializationException;
 import com.google.gwt.user.client.rpc.SerializationStreamReader;
 import com.google.gwt.user.client.rpc.SerializationStreamWriter;
 
 @${generated}("com.google.auto.value.processor.AutoValueProcessor")
-public final class $serializerClass {
-  // There are no instances of this class.
-  private ${serializerClass}() {}
+public final class $serializerClass$formalTypes extends CustomFieldSerializer<$subclass$actualTypes> {
```
eamonnmcmanus: Could you explain why you needed to change this?

tbroyer: It's an additional (optional) enhancement to reduce the amount of reflection at runtime on server side.

When GWT-RPC finds a custom field deserializer class, if it extends `CustomFieldSerializer`, it'll instantiate it and then call the instance methods; otherwise it'll use reflection to get the static methods and call them. So extending `CustomFieldSerializer` reduce the amount of reflection needed.
There's also a `ServerCustomFieldSerializer` for _type-checked_ deserialization; IIUC this is mostly needed for generic classes, so that could be another enhancement step for auto-value.

See http://www.gwtproject.org/javadoc/latest/com/google/gwt/user/client/rpc/CustomFieldSerializer.html and http://www.gwtproject.org/doc/latest/DevGuideServerCommunication.html#DevGuideCustomSerialization

value/src/it/gwtserializer/src/test/java/com/google/auto/value/client/GwtSerializerTest.java
tbroyer: FYI, that test was guided by https://stackoverflow.com/q/31221158/116472
I was looking at how that was implemented and saw there was no test, so I wrote one and specifically tested nested autovalue types. Looking at the code, I noticed support for generics, so I also tested them.
And once that was done, I enhanced serialization by extending `CustomFieldSerializer`.

eamonnmcmanus:
As I recall, when I added GWT serialization support, I originally did have both the static and the instance methods. When I realized that the instance methods weren't necessary I deleted them, since it annoyed me to have two sets of methods to do the same thing. If it were possible to delete the static methods instead that would be idea, but for some reason that doesn't work.

I'm not sure we really care all that much about the reflection cost on the server, which will surely be dwarfed by the other costs in any GWT interaction. We don't necessarily care very much about three tiny extra methods per serializer, but still. Do you think there are other good reasons to add the instance methods? Perhaps the `ServerCustomFieldSerializer` you mentioned? Though generic classes seem to work fine right now.

tbroyer:
> I'm not sure we really care all that much about the reflection cost on the server, which will surely be dwarfed by the other costs in any GWT interaction. We don't necessarily care very much about three tiny extra methods per serializer, but still.

The `CustomFieldSerializer` class was added 4 years ago in a commit titled “Optimizations for server-side invocations of CustomFieldSerializers.” so I assumed the code had been profiled and this was a bottleneck, and the optimization had been benchmarked; though maybe I'm wrong.
…and JVMs have evolved since then too, so maybe it's no longer an issue.

> Do you think there are other good reasons to add the instance methods? Perhaps the `ServerCustomFieldSerializer` you mentioned? Though generic classes seem to work fine right now.

FYI, the `ServerCustomFieldSerializer` has been added 7 months after the `CustomFieldSerializer` in a commit whose commit message was “Adds pre-deserialization type checking for values in GWT RPC messages. The server will now avoid deserializing types that mismatch the RPC method that is being invoked, reducing the chance that an erroneous message will cause uncaught deserialization errors on the server.”
So indeed generic classes do work without it, it'd “just” be an added safety net.

I don't have numbers to argue one way or another, so if you think it's not worth it, I can just remove the commit from the branch to leave only the integration tests.

eamonnmcmanus:
The GWT expert who helped me develop this code is @skybrian. Brian, is it worth changing AutoValue so that `Foo_CustomFieldSerializer` classes extend `CustomFieldSerializer`, and implement its methods to forward to the existing static methods?

skybrian:
Unfortunately I don't have any performance metrics either. Not a whole lot to add; I agree with what was said so far, and Thomas Broyer is as much a GWT expert as me :-)

I'd be inclined to do it for the theoretical performance gain. It could reduce CPU usage on servers that handle lots of GWT-RPC traffic and don't do a whole lot else. (A request or response might contain a large array of pair objects, for example.) The downside is a few extra forwarding methods per generated class, which doesn't seem like a big deal on a server. It would be on Android, but I'd guess they're unlikely to use GWT-RPC and won't use this generated class?

Perhaps some Adwords folks could check the performance?

eamonnmcmanus:
OK. I just want to run the change against Google's code base to make sure it doesn't break anything. I'll merge the PR later today if all goes well.

eamonnmcmanus:
OK, I ran the template change against the complete set of Google's internal tests that are transitively affected by it and there were no obvious problems, so I think we are OK.

PR #259, opened 2015-08-05
Merge in changes from internal google repository.

cgruber:
Substantial changes include: 

### AutoValue

  - Replacement of Velocity with a home-grown alternative by @eamonnmcmanus called "escape velocity"
    * Eases issues with shading, as well as avoiding some workarounds necessary with apache's velocity.
  - Fixed nullable propagation

### Fixes to AutoFactory

  - Qualifier annotations for parameters
  - primitive type fixes
  - better propagation of annotations in certain cases
  - naming collision handling.

### Common

  - A more robust approach to element validation in BasicAnnotationProcessor
  - Handle method parameter annotations in  BasicAnnotationProcessor
 
```
NOTE: This is a breaking change for users of Common.
```

### Various minor build and other fixes. 

  - build speed bump on ubuntu
  - Remove some type inferencing that breaks earlier versions of javac.


cgruber:
Manually fixed CLA issues since all authored changes in this commit are from googlers.

eamonnmcmanus:
LGTM. Waiting for the checks to complete before merging, though.

eamonnmcmanus:
Look as if this doesn't build, because I moved a method from value/ to common/, so I guess value/pom.xml needs to reference a snapshot of common?

cgruber:
Yeah - And common now uses features in compile-testing 0.7.

cgruber:
At this point I'm confused.  This buidls on my mac, it builds on ubuntu, but it fails in travis because of this:

/usr/local/maven-3.2.5/bin/mvn: 3: export: -Xmx512m: bad variable name

I feel like this is a travis issue, not an issue here, and am inclined to push it in anyway... except that it then won't produce snapshots. :(  I'll try to trouble shoot it and figure out what's wrong here.

eamonnmcmanus:
Looks like https://github.com/travis-ci/travis-ci/issues/4621. Hopefully it will be fixed soon. I agree that we should merge anyway.

cgruber:
Ok.  Merging, and I'm manually deploying snapshots.

PR #260, opened 2015-08-07
Stop cloning arrays in @AutoValue classes. 

cgruber:
Some projects have been bitten by the unexpected cost of the array-cloning that was happening whenever a primitive-array-valued property of an @AutoValue class was read. Others chimed in to confirm that they were surprised to learn from the discussion that that was what happened. 

A weak defense was made in favor of array copying as necessary to guarantee immutability but the consensus inside google was that this was not a strong argument.

If users of autovalue want their array-valued property to be strictly immutable (because they don't trust the clients of their class not to mutate the array) then arrays can be wrapped in an immutable class, such as the ones provided by fastutil.

-------------
Created by MOE: http://code.google.com/p/moe-java
MOE_MIGRATED_REVID=99968654


JakeWharton:
Can the rest of us in the world get a summary or something of the points made in the discussion?

eamonnmcmanus:
We had a couple of people who were bitten independently by the unexpected cost of the array-cloning that was happening whenever you read a primitive-array-valued property of an `@AutoValue` class. Others chimed in to confirm that they were surprised to learn from the discussion that that was what happened. I weakly defended array copying as necessary to guarantee immutability but the consensus was that this was not a strong argument. If you want your array-valued property to be strictly immutable (because you don't trust the clients of your class not to mutate the array) then you can wrap the array in an immutable class, such as the ones provided by [fastutil](http://fastutil.di.unimi.it/).

cgruber:
I've updated the comment with Eamon's commentary, edited for flow.

cgruber:
Other than clabot not being able to process Eamonn, this commit passes, so I'm going to merge it.

JakeWharton:
Makes sense. I fall in the always-guard camp as well, but I can see how the
potentially-hidden nature of the behavior might surprise (and have serious
effects).

PR #262, opened 2015-08-07
Create NOTICE of apache license

cgruber:
This should have been there, but wasn't synced-out properly earlier.


eamonnmcmanus:
I don't think we need to wait for Travis before merging this.

PR #264, opened 2015-08-10
As suggested by @jbgi, change the Maven build for AutoValue to use the org.immutables.tools...

eamonnmcmanus:
...version of the Maven Shade Plugin, which supports prefixing the simple names of shaded classes with $. This means that people can safely use the AutoValue with their IDEs, without having code completion ever show the shaded classes.


eamonnmcmanus:
This change was merged via Google's code base.

PR #265, opened 2015-08-12
Adds support for Extension property consumption.

rharter:
Adds ability for extensions to indicate that they intend to consume properties so that AutoValue doesn't generate methods for them, or include them in builders and other generated methods.

The AutoValueExtension interface is updated to include a `consumeProperties` method that will return a collection of property names from the `context.properties()` map that it intends to implement.  

My initial implementation of Extensions with the reversed inheritance provided for this by removing properties from the map during class generation, but since the AutoValue class is the first to be generated now, this had to be done in a separate pass.


value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -397,6 +387,26 @@ private void processType(TypeElement type) {
       }
     }
 
+    if (appliedExtensions.size() > 0) {
+      final Set<String> methodsToRemove = Sets.newHashSet();
+      for (int i = appliedExtensions.size() - 1; i >= 0; i--) {
+        AutoValueExtension extension = appliedExtensions.get(i);
+        Collection<String> consumed = extension.consumeProperties(context);
+        if (consumed != null) {
```
JakeWharton: `null` collections make me :cry: You're also using it in the test. Perhaps make `AutoValueExtension` an abstract class with a default implementation that returns `Collections.emptySet()` and forbid `null`?

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -434,26 +444,24 @@ private void processType(TypeElement type) {
...
+    ImmutableSet.Builder<T> builder = ImmutableSet.builder();
+    for (T item : original) {
+      if (predicate.apply(item)) builder.add(item);
+    }
+    return builder.build();
+  }
 
-      @Override public Map<String, ExecutableElement> properties() {
-        return propertyNameToMethodMap(methods);
+  private <K, V> ImmutableBiMap<K, V> newImmutableBiMapRemovingKeys(ImmutableBiMap<K, V> original,
```
JakeWharton: This whole method is just
```java
original.keySet().removeAll(keysToRemove)
```

JakeWharton: Ugh, nevermind. **Immutable**BiMap

rharter: Was just going to point that out.

value/src/main/java/com/google/auto/value/processor/ExtensionContext.java
```diff
@@ -0,0 +1,49 @@
...
+  public ExtensionContext(ProcessingEnvironment processingEnvironment, TypeElement typeElement,
+                          Map<String, ExecutableElement> properties) {
+    this.processingEnvironment = processingEnvironment;
+    this.typeElement = typeElement;
+    this.properties = properties;
+  }
+
+  @Override
+  public ProcessingEnvironment processingEnvironment() {
+    return processingEnvironment();
```
JakeWharton: remove `()`

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
JakeWharton: I almost made the same mistake with `newFilteredImmutableSet` and `Sets.filter`...

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
rharter: Originally I would going to make the method `@NonNull Set<String> consumeProperties(Context context)`, would that suffice?  I am hesitant to convert this to a class simply for that.

rharter: Also, the test intentionally returns null since it's possible based on the interface and I need to test that case.

rharter:
@eamonnmcmanus Travis seems to be failing on a test that is untouched in this PR.  Is that on me to fix for this?

eamonnmcmanus:
I'm afraid I won't be able to get this for a couple of weeks. But I do have a question:

> My initial implementation of Extensions with the reversed inheritance provided for this by removing properties from the map during class generation, but since the AutoValue class is the first to be generated now, this had to be done in a separate pass.

Is that a fundamental restriction? It seems as if it would be a lot simpler if extensions were called before the generation of the AutoValue implementation, and allowed to remove entries from the properties map.

rharter:
@eamonnmcmanus You bring up a good point.  I was thinking that it would need to generate the Auto implementation first so that the extensions have something to extend (same flow as when I manually write code), but since this happens in one pass without trying to compile the generated code in between there's no reason we can't flip it around and generate the subclasses first, then the Auto implementation with the resulting properties.

One stipulation would be that each non-final extension *must* create a constructor taking all properties passed into it, so it's subclasses (which won't have knowledge of it's consumed properties) won't call `super()` with arguments that aren't accepted.

Another issue that this might introduce is the, currently, the properties are an ImmutableMap, so order is guaranteed (important for constructor arg lists), but making that list mutable could introduce complexity around that.  Do you think that is a warranted concern?

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
eamonnmcmanus: I think it would be fine to require extensions to return a non-null set, which as @JakeWharton notes could be `Collections.emptySet()`. I don't think it's much of a burden for them to have to do that explicitly, though probably we should have `AutoValueExtension` be an abstract class so we can add methods to it (without using Java 8 default methods), and in that case obviously this could have a default implementation as described.

eamonnmcmanus:
The point about constructors is a good one. I think that does warrant having a separate pass to allow properties to be claimed, so that they don't appear in any constructor. That does mean that we have three passes all told though: one where the extension says whether it is applicable, one where it can claim properties, and one where it generates code. I wonder if we couldn't simplify that somehow.

rharter:
I agree about it being a large number of passes.  If you don't think order is important we could combine the first two passes:

```java
for (Extension e : extensions) {
  if (e.applicable(context)) {
    appliedExtensions.add(e);
    propertiesToRemove.addAll(e.consumeProperties(properties));
  }
}
```

It still ends up calling into the extension twice, but only loops once.  Given that this is compile time instead of runtime do you think that's an important change that would provide some meaningful benefit?

eamonnmcmanus:
I was more concerned with simplifying the conceptual model than the implementation. I'm not seeing a good way to do that right now, though.

rharter:
I've updated this to document that `consumeProperties()` returns a non-null value and changed AutoValueExtension to an abstract class so we can provide default implementation.

value/src/main/java/com/google/auto/value/AutoValueExtension.java
```diff
@@ -19,13 +20,14 @@
  * generated class. Its not guaranteed that an Extension's generated class will be the final
  * class in the inheritance hierarchy, unless it's {@link #mustBeAtEnd} method returns true,
```
JakeWharton: This reference needs updated.

eamonnmcmanus: I'm not sure I see what's wrong with it?

rharter: The method has been updated to `mustBeFinal`.

JakeWharton: Yeah. Doclint. Worst JDK change ever?

eamonnmcmanus:
We've moved AutoValueExtension to its own package, so that change will need to be merged here. Hopefully that will be straightforward.

rharter:
Updated

eamonnmcmanus:
Sorry to ask again, but can you resync? It should be trivial I think.

cgruber:
Hi.  To clarify, please rebase this branch.  We'd prefer it if you squashed the branch into a single commit also.

rharter:
rebased and squashed.

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -638,18 +645,13 @@ private static void determineObjectMethodsToGenerate(
...
       TypeElement autoValueClass, Set<ExecutableElement> methods) {
     ImmutableSet.Builder<ExecutableElement> toImplement = ImmutableSet.builder();
-    Set<Name> toImplementNames = Sets.newHashSet();
     boolean ok = true;
     for (ExecutableElement method : methods) {
       if (method.getModifiers().contains(Modifier.ABSTRACT)
           && objectMethodToOverride(method) == ObjectMethodToOverride.NONE) {
         if (method.getParameters().isEmpty() && method.getReturnType().getKind() != TypeKind.VOID) {
           ok &= checkReturnType(autoValueClass, method);
-          if (toImplementNames.add(method.getSimpleName())) {
```
eamonnmcmanus: This looks like a merge error. This code should not have been deleted.

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -410,26 +427,24 @@ private void processType(TypeElement type) {
...
-
-      @Override public String packageName() {
-        return TypeSimplifier.packageNameOf(type);
-      }
-
-      @Override public TypeElement autoValueClass() {
-        return type;
-      }
+  private <T> ImmutableSet<T> newFilteredImmutableSet(ImmutableSet<T> original,
+                                                      Predicate<T> predicate) {
```
felipecsl: the formatting is off here, not matching the rest of the project (disable "Align when multiline" for "Method call arguments" in your editor code style).

felipecsl: also looks like this method can be static

eamonnmcmanus: Yes. At this point, though, I've pulled the code into the Google machine and I'd already fixed that particular issue. I'll be making some more changes before it comes out again. :-/

rharter:
@eamonnmcmanus Any update on this?

eamonnmcmanus:
I will try to get to it next week. Bug me if there's no progress in a week.

rharter:
@eamonnmcmanus Consider yourself bugged :)

eamonnmcmanus:
I've tweaked this code and am getting it reviewed before submitting it to Google's internal repo. So there is progress. :-)

cgruber:
There.  The format of the merge is weird, because I created a merge CL so your version and our version are both merged together, which then went into a different pull that merged all our exports, but the effect is... your version and our version are both credited to you and close this pull request, which is marked as merged, though our internal version is canonical.  But the changes are largely formatting and some minor tweaks.

PR #279, opened 2015-10-22
Merge from writing branch from 4a65dd332091df2512bcc24fffd5a4fdcc64119c

cgruber:
* Merges in P/R #265 (with google edits, attribution preserved)
  * fixes some documentation
  * unifies guava version info in the pom file
  * explicitly causes compiler errors rather than poorly generated code for non-static nested A/V types
  * ports a test to JUnit 4.


cgruber:
clabot lies.

PR #285, opened 2015-11-12
Fixes javadoc reference errors after refactor.

rharter:
The refactor of the `extension` package left some references in the javadocs broken.

### DIFF
--- a/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
+++ b/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
@@ -20,14 +20,16 @@ import javax.lang.model.element.TypeElement;
  *
  * <p>Extensions can extend the AutoValue implementation by generating subclasses of the AutoValue
  * generated class. It is not guaranteed that an Extension's generated class will be the final
- * class in the inheritance hierarchy, unless its {@link #mustBeFinal()} method returns true. Only
- * one Extension can return true for a given context. Only generated classes that will be the
+ * class in the inheritance hierarchy, unless its
+ * {@link com.google.auto.value.extension.AutoValueExtension#mustBeFinal(Context)} method returns true.
+ * Only one Extension can return true for a given context. Only generated classes that will be the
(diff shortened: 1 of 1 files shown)


value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
```diff
@@ -82,8 +84,8 @@ public boolean applicable(Context context) {
    * in the inheritance hierarchy.  Only one extension may be the final class, so
    * this should be used sparingly.
    *
-   * @param context The {@link com.google.auto.value.AutoValueExtension.Context} of the code
-   *                generation for this class.
+   * @param context The {@link com.google.auto.value.extension.AutoValueExtension.Context} of the
```
eamonnmcmanus: I think it's enough to write Context. There will be a link to that type in the javadoc anyway since it's the type of the parameter.

value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
```diff
@@ -91,10 +93,11 @@ public boolean mustBeFinal(Context context) {
   }
 
   /**
-   * Returns a non-null set of property names from {@link #properties()} that this extension intends
-   * to implement.  This will prevent AutoValue from generating an implementation, and remove the
-   * supplied properties from builders, constructors, {@code toString}, {@code equals},
-   * and {@code hashCode}. The default set returned by this method is empty.
+   * Returns a non-null set of property names from {@link AutoValueExtension.Context#properties()}
```
eamonnmcmanus: Here {@link Context#properties()} is enough.

eamonnmcmanus:
Thanks! I'll apply changes along those lines to Google's internal copy and they should appear at the next push.

rharter:
Just to be clear, are you saying I should not worry about making the changed you outlined above? You'll take care of it in internal and it'll come out in the next push?

eamonnmcmanus:
Yes. I'm going to close this PR but it will be referenced in the commit log of the next push. Thanks for fixing these problems!

Issue #287, opened 2015-11-18
can not have create and a builder at the same time.


hgschmie:
this results in the create static method trying to call the `AutoValue_Foo` constructor which is now private because a builder is present.

the constructor should only be private if only a builder is present, but package private if other create methods are present.

Our use case is that there are places where a bean is created incrementally (from business logic code) and places where we want an uniform "all at once" approach is wanted (deserialization from JSON e.g.). 

The create method could use the builder but this requires a large amount of not-autogenerated boilerplate code.

eamonnmcmanus:
AutoValue can't tell if there is a static factory method. It doesn't require such methods to be called `create`, and it can't look inside the code of a method to see whether the method calls the `AutoValue_Foo` constructor. Having said that, I'm not sure there's a really strong reason for the constructor to be private when a builder is present. Maybe it could be package-private whether or not there is a builder. Comments invited from @kevinb9n. :-)

kevinb9n:
It's extremely important that AutoValue be extremely easy to learn and
reason about, and I'm a little wary of confusion we could be introducing if
we do this. As just one example: what will users assume about whether the
generated constructor is calling their validation method or not?  Perhaps
we can deal with that, but I suspect other little wrinkles like that could
be hiding too.

On Wed, Nov 18, 2015 at 2:41 PM, Éamonn McManus <notifications@github.com>
wrote:

> AutoValue can't tell if there is a static factory method. It doesn't
> require such methods to be called create, and it can't look inside the
> code of a method to see whether the method calls the AutoValue_Foo
> constructor. Having said that, I'm not sure there's a really strong reason
> for the constructor to be private when a builder is present. Maybe it could
> be package-private whether or not there is a builder. Comments invited from
> @kevinb9n <https://github.com/kevinb9n>. :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/auto/issues/287#issuecomment-157888399>.
>



-- 
Kevin Bourrillion | Java Librarian | Google, Inc. | kevinb@google.com

hgschmie:
@kevinb9n I understand the desire for simplicity. Right now, I can not have an explicit constructor (which would force callers to provide values for all the fields in the bean in an atomic fashion) *and* a build that would allow incremental construction and validation. So I resorted to having my create methods to use the builder which requires me to write the boilerplate code that I want to avoid.

Making the c'tor non-private would solve my problems very nicely. :-)

hgschmie:
[...]

PR #289, opened 2015-11-23
Makes autovalue bean constructor always visible

hgschmie:
Changes the visibility of the constructor for the generated bean to be
unconditionally package private.

Now it is possible to have static construction methods and a builder
at the same time e.g. when creating a bean incrementally in logic
while deserializing it wholesale from e.g. JSON or Protobuf.


hgschmie:
This is a possible fix for #287.

PaulWoitaschek:
I'd like to see this happen as Im using both for performance reasons.

eamonnmcmanus:
We don't want to do this unconditionally for the reasons mentioned in the discussion on #287.

Issue #283, opened 2015-11-04
AutoValue should respect @Nullable annotations on types


philipp-classen:
AutoValue generates non-null runtime checks for all fields that are not annoted as "@Nullable". At the moment, "@Nullable" annotations using  JSR 308 type annotations are ignored. Thus, non-null checks will be generated for fields which are intended to be nullable.

---

For example, the Eclipse "@Nullable" annotation looks like this:

```
@Documented
@Retention(value=CLASS)
@Target(value=TYPE_USE)
public @interface Nullable
```
Source: http://help.eclipse.org/mars/index.jsp?topic=/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/annotation/Nullable.html

Note that it only defines "TYPE_USE" as Target. It is a new feature that is available since Java 8.

Before that, it was defined as follows:
```@Target({ ElementType.FIELD, ElementType.LOCAL_VARIABLE, ElementType.METHOD, ElementType.PARAMETER })```

AutoValue handles the old annotation fine, but ignores the new TYPE_USE based annotation. Consequently, it generates non-null runtime checks for all non-annotated fields, it is critical that detects "@Nullable" annotation.

Note: Mixing TYPE_USE and the older annotations (FIELD, LOCAL_VARIABLE, METHOD, PARAMETER) is not a good idea. Though it works in simple examples, it introduces ambiguity, for instance, when using it with arrays. Thus, Eclipse rightly decided against it.

philipp-classen:
My colleague has an experimental patch to add support for type annotations. It uses the Java 8 API, but, in principle, it could be implemented in a Java 7 backward compatible way (using reflection).

We hope to find time to provide a pull request, so you can take a look at the idea.

eamonnmcmanus:
Yes, I have an internal issue tracking this. As you point out, the need to have the code work on earlier Java versions means it isn't trivial, but it shouldn't be all that hard either.

philipp-classen:
Please take a look at the patch in pull request #293

eamonnmcmanus:
The patch in #293 has been incorporated into the latest snapshot.

sudhirj:
Has this regressed in 1.5 and 1.5.1? I can't get `@Nullable` to work at all. Null checks are always being generated.

eamonnmcmanus:
There are no differences between 1.5 and 1.5.1 that could explain that. Are you sure that is the only thing that has changed? Does it work using 1.5 and JDK9?

[...]

PR #293, opened 2015-11-30
patch for "AutoValue should respect @Nullable annotations on types #283"

brychcy:
The patch invokes the required java8-method via reflection, so everything should work with earlier java version (verified with java 7). The tests for java 8 functionality are ignored with earlier java versions, too.

What currently doesn't work (because of javac limitations):
- @Nullable primitive arrays (I added an @Ignored test case for this)
- @Nullable in type annotation position, e.g. 
abstract java.lang.@Nullable String anString()

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ script:
 jdk:
   - oraclejdk7
   - openjdk7
+  - oraclejdk8
 
 branches:
   except:
--- a/value/src/it/functional-java8/invoker.properties
+++ b/value/src/it/functional-java8/invoker.properties
@@ -1,0 +1,6 @@
+invoker.goals = clean verify
+
+invoker.profiles.1 =
+invoker.profiles.2 = eclipse
+
+invoker.java.version=1.8+
--- a/value/src/it/functional-java8/pom.xml
+++ b/value/src/it/functional-java8/pom.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright (C) 2013 Google, Inc.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<!-- TODO(gak): see if we can manage these dependencies any better -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.auto.value.it.functional</groupId>
+  <artifactId>functional-java8</artifactId>
+  <version>HEAD-SNAPSHOT</version>
+  <name>Auto-Value Functional Integration Test (Java 8)</name>
+  <dependencies>
+    <dependency>
+      <groupId>com.google.auto.value</groupId>
+      <artifactId>auto-value</artifactId>
+      <version>@auto.version@</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>0.25</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava-testlib</artifactId>
+      <version>18.0</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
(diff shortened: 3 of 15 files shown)


eamonnmcmanus:
Thanks for this contribution! If I could make a couple of requests:

* We use two-space indentation at Google rather than four.
* Rather than having a Java 8 version of CompilationTest, I think it would be more maintainable to have a Java 8 version of AutoValueTest. I'd like to minimize the amount of code that tests the exact code generated, because it's a pain to update it all whenever we change the details of code generation.

Since the main motivation for this change is to ensure that null values are allowed if a property is `@Nullable` it should be straightforward to test that that is happening even for a type-targeted `@Nullable`.

FYI the way our code flow works we won't be merging this PR. Instead, we'll submit it into Google's internal code base and it will appear at our next push as a later PR that references this one. We do need the CLA to be signed before we can start that.

brychcy:
"Rather than having a Java 8 version of CompilationTest, I think it would be more maintainable to have a Java 8 version of AutoValueTest. I'd like to minimize the amount of code that tests the exact code generated, because it's a pain to update it all whenever we change the details of code generation."

I felt a bit of the pain while creating the test, so I really understand what you mean. The problem is that TYPE_USE annotations don't compile at all with java 7 or older, so the integration tests wouldn't even compile and I currently have no idea how to exclude the tests on the maven level depending on the jdk.

I'll have a look at the indentation and CLA.

tbroyer:
Wrt the test, you'd create a new integration-test project (e.g. `src/it/functional-java8`) and put an `invoker.properties` in there with `invoker.java.version=1.8+`. That way the integration-test project would only ever be invoked when the build is done with a JDK 8.
See https://maven.apache.org/plugins/maven-invoker-plugin/examples/selector-conditions.html
You'd also add a JDK 8 to the `.travis.yml`.

There probably are ways to get it run with Java 8 while Maven is invoked with a JDK 7, using Maven Toolchains, but AFAICT it's a PITA to get it working and would require everyone willing to build the project to configure his Maven toolchains.

brychcy:
I signed it!

brychcy:
I signed it!

brychcy:
Thanks for the hints regarding maven, I'll try to create a java 8 version of AutoValueTest.

brychcy:
It looks like AutoValueTest currently doesn't get executed at all.

The files  `value/src/it/functional/invoker.properties ` and  `value/src/it/functional/pom.xml` were deleted in commit https://github.com/google/auto/commit/2c4bf2255c535ea7746eebab466d0026a3503c34 

Maybe this was not intended?

eamonnmcmanus:
Yow! It looks as if something pretty bad happened in that commit. We'll need to investigate. (Paging @cgruber...)

cgruber:
Fixed.  Please fetch the remote origin branch and rebase this pull
requests' branch against origin/master.  For some reason a README/javadoc
editing commit got mis-merged (without conflict) and the mis-merge deleted
the two files needed to invoke the integration tests.  That is crazy weird,
but I've fixed it by partially reverting that commit (just the relevant
bits).  It seems likely it was bad behavior from our syncing tool, and we
missed it.  Apologies.

brychcy:
Updated.

Some notes:

1) The "eclipse" profile in value/src/it/functional/pom.xml doesn't really seem to do anything different than the other profile. Support for com.google.auto.value.EclipseHackTest has been removed in 4b8ccc5dc742631936987ec836107325ec50d706. The commit comment suggests that there is a real test with the Eclipse compiler, but i cannot see how it works (maybe Google-internal somehow?). I have therefore manually tested with eclipse (and have added some code to TypeSimplifier)

2) I have copied functional to functional-java8 as suggested. You can compare the directories, to see what had to be changed. Note that I have disabled the NullPointerTester tests (because it doesn't understand type annotations) and some tests in the functional-java8 version of AutoValueTest.java because of the following limitations (repeated from the commit comment):

Limitations: The following cases currently don't work with javac (i.e.
the annotation processors don't see the @Nullable so a null check is
generated in the constructor):

public abstract int @Nullable [] getInts()
This works with eclipse JDT and the null check is ommited in the
generated code, but note that the @Nullable is in the wrong position in
the generated code (i.e. @Nullable int[] instead of int @Nullable[])),
which might lead to compiler warnings

For generic type T: 
@Nullable
public abstract T getT()
(works in JDT)

Also @Nullable won't work for types which can't be imported because of
naming conflicts, because the @Nullable is generated at the wrong place,
i.e. "@Nullable some.package.Map" instead of "some.package.@Nullable
Map" which won't compile at all for TYPE_USE @Nullable.

brychcy:
Regarding the javac limitations: openjdk bug [8031744](https://bugs.openjdk.java.net/browse/JDK-8031744) is fixed for java 9 as of [JEP 217] (http://openjdk.java.net/jeps/217). 

I have tried the it/functional-java8 test with the [Early Access Java 9 Build 95](https://jdk9.java.net/download/) and the annotations are now visible to the Annotations processor, so the `int @Nullable []` case works. What still doesn't work is `@Nullable T`, because for it, the `TypeMirror#toString()` which is used via `com.google.auto.value.processor.TypeSimplifier.ToStringTypeVisitor.defaultAction(TypeMirror, StringBuilder)` now includes the Annotation, so the generated code contains the Annotations twice:
```java
@Nullable
@Override
public @com.google.auto.value.annotation.Nullable T t() {
    return t;
}
```

eamonnmcmanus:
Thanks for this detailed investigation!

I'm not sure whether JDK 9 should be including annotations in the toString. Maybe. Regardless, avoiding the problem seems straightforward, by overriding `visitTypeVariable` in `TypeSimplifier.ToStringTypeVisitor`, and using `typeVariable.asElement().getSimpleName()`.

This makes me think that `ToStringTypeVisitor` should document exactly where it is used so that it will be more obvious what string should be output for each kind of type. But that's an issue for me.

eamonnmcmanus:
This code has been incorporated into the latest snapshot. Thanks!

Our processes mean that we do this in Google's internal repository rather than merging the code here, so I'm closing the PR.

PR #299, opened 2015-12-30
Moves method verification after extensions have had the chance to consume properties.

rharter:
Since compile-testing [doesn't support](https://github.com/google/compile-testing/issues/41) checking warnings, I've simply tested for the inverse here, ensuring that invalid properties fail with the appropriate error.  This move should also remove the warning for "abstract methods other than property getters and builder converters" for consumed methods.

### DIFF
--- a/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
+++ b/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
@@ -18,7 +18,8 @@ import javax.lang.model.element.TypeElement;
  *
  * <p>Extensions can extend the AutoValue implementation by generating subclasses of the AutoValue
  * generated class. It is not guaranteed that an Extension's generated class will be the final
- * class in the inheritance hierarchy, unless its {@link #mustBeFinal(Context)} method returns true.
+ * class in the inheritance hierarchy, unless its
+ * {@link com.google.auto.value.extension.AutoValueExtension#mustBeFinal(Context)} method returns true.
  * Only one Extension can return true for a given context. Only generated classes that will be the
  * final class in the inheritance hierarchy can be declared final. All others should be declared
  * abstract.
@@ -81,7 +82,8 @@ public abstract class AutoValueExtension {
    * in the inheritance hierarchy.  Only one extension may be the final class, so
    * this should be used sparingly.
    *
-   * @param context The Context of the code generation for this class.
+   * @param context The {@link com.google.auto.value.extension.AutoValueExtension.Context} of the
+   *                code generation for this class.
    * @return True if the resulting class must be the final class in the inheritance hierarchy.
    */
   public boolean mustBeFinal(Context context) {
@@ -89,10 +91,11 @@ public abstract class AutoValueExtension {
   }
 
   /**
-   * Returns a non-null set of property names from {@link Context#properties()} that this extension
-   * intends to implement. This will prevent AutoValue from generating an implementation, and remove
-   * the supplied properties from builders, constructors, {@code toString}, {@code equals},
-   * and {@code hashCode}. The default set returned by this method is empty.
+   * Returns a non-null set of property names from {@link AutoValueExtension.Context#properties()}
+   * that this extension intends to implement.  This will prevent AutoValue from generating an
+   * implementation, and remove the supplied properties from builders, constructors,
+   * {@code toString}, {@code equals}, and {@code hashCode}. The default set returned by this
+   * method is empty.
    *
    * <p>For example, Android's {@code Parcelable} interface includes a
    * <a href="http://developer.android.com/reference/android/os/Parcelable.html#describeContents()">method</a>
@@ -105,7 +108,8 @@ public abstract class AutoValueExtension {
    * implementation and return a set containing {@code "describeContents"}. Then
    * {@code describeContents} will be omitted from builders and the rest.
    *
-   * @param context The Context of the code generation for this class.
+   * @param context The {@link com.google.auto.value.extension.AutoValueExtension.Context} of the code
+   *     generation for this class.
    * @return A collection of property names that this extension intends to implement.
    */
   public Set<String> consumeProperties(Context context) {
--- a/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
+++ b/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
@@ -18,16 +18,14 @@ import javax.lang.model.element.TypeElement;
  *
  * <p>Extensions can extend the AutoValue implementation by generating subclasses of the AutoValue
  * generated class. It is not guaranteed that an Extension's generated class will be the final
- * class in the inheritance hierarchy, unless its
- * {@link com.google.auto.value.extension.AutoValueExtension#mustBeFinal(Context)} method returns true.
+ * class in the inheritance hierarchy, unless its {@link #mustBeFinal(Context)} method returns true.
  * Only one Extension can return true for a given context. Only generated classes that will be the
  * final class in the inheritance hierarchy can be declared final. All others should be declared
  * abstract.
  *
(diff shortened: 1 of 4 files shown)


rharter:
@eamonnmcmanus This is failing the build on OpenJDK in some unrelated tests, with a buffer overflow in libnet.so.  Any idea what that's all about?

tbroyer:
@rharter I've seen crashes during tests on Travis CI with OpenJDK 7 too: https://travis-ci.org/tbroyer/gradle-errorprone-plugin/builds/99365020
Switched from the "legacy precise" to the container-based infra and it went back to green: https://travis-ci.org/tbroyer/gradle-errorprone-plugin/builds/99366548

cgruber:
We have some confirmation that this is a bug in the JDK which was fixed in 8 and backported to Oracle's JDK7, but not into OpenJDK7.  I'm looking at a different way to fix this, as a different htmlunit runstyle could override things, but that would involve a new custom library just to run gwt tests via htmlunit on OpenJDK7, and given the real-world prevalence of 8, the non-crashiness of Oracle's jdk 7, and the container workaround I think I'm split between pushing for container use on Travis, or just dropping OpenJDK7 from the test rig (or disabling gwt tests on that jvm)

cgruber:
Note - we're having this issue on Truth now too.

cgruber:
@rharter - I believe warnings are in, or about to be in Compile-Testing, and i'll sync it out soon.

rharter:
@cgruber Great, I'll look into adding warning tests once that's done.

cgruber:
Crap... I found that that change (warnings API on compile-testing) is still pending... but "real soon now" (next day or two, just resolving some final review issues).

rharter:
@cgruber Did this ever get merged?  I don't see anything in the `JavaSourcesSubject` that would indicate warning validation.

cgruber:
Hmm.  I was under the impression that it had, and that we'd synced out more recently than it seems we have... I know it went in internally, but apparently we're behind on syncing out.  Let me check on what's going on. :/

cgruber:
Yep.  Sorry, it needs a sync, and I'll make sure we get a SNAPSHOT out shortly (this week), and probably a stable minor release soon after.

rharter:
That sounds great.  Thanks!

eamonnmcmanus:
@cgruber I don't think this change has been pulled in to the Google version.

rharter:
@cgruber It looks like the changes have been merged into compile testing. Do you currently have a plan for a new maven release?  I can't use this it here until you release it.

rharter:
@cgruber Any update on this? ^^

JakeWharton:
@rharter Looks like compile-testing 0.9 is out

rharter:
nice, I'll get this updated.

rharter:
I've tried updating the tests to use the new `compilesWithoutWarnings()` method from compile-testing, but we actually always compile **with** at least one warning, `No processor claimed any of these annotations: com.google.auto.value.AutoValue`.  This is apparently intentional as documented in the [code](https://github.com/google/auto/blob/master/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java#L161).  The `compilesWithoutWarnings` method doesn't allow me to access the actual warnings to filter out known, expected warnings, so I can't filter out that warning and ensure it doesn't contain extra warnings.

I think this means that the current negative test, checking that the compilation fails for unconsumed properties will have to suffice.

eamonnmcmanus:
If you know that there will always be this one warning, could you use `withWarningCount(1)`?

rharter:
Ahh, didn't realize that that was added too, I'll look into that.

tbroyer:
Is it possible to pass `-Xlint:-processing`? That would remove that warning.

rharter:
@eamonnmcmanus I believe this is ready to go, now.

menny:
I would LOVE to see this in!
Our project has `-Werror` compiler flag, and without this PR in, we can not use AutoValue (and auto-value-parcel) in our source code.

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
```diff
@@ -120,12 +121,49 @@ public void testExtensionConsumesProperties() throws Exception {
...
+        "@AutoValue public abstract class Baz {",
+        "  abstract String foo();",
+        "  abstract String dizzle();",
+        "}");
+    assertAbout(javaSource())
+        .that(impl)
+        .processedWith(new AutoValueProcessor(ImmutableList.of(new FooExtension())))
+        .compilesWithoutError()
+        // expected warnings: @AutoValue and @Generated annotations not consumed
+        .withWarningCount(2);
```
eamonnmcmanus: I've had success passing `.withCompilerOptions("-Xlint:-processing")` to turn off the warnings about `@AutoValue` and `@Generated` not being consumed by any processor. However, when I do that in this test, it passes without any other changes to the code. I think we need a test that fails before the rest of the modifications in this PR are made and passes afterwards.

value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
```diff
@@ -18,7 +18,8 @@
  *
  * <p>Extensions can extend the AutoValue implementation by generating subclasses of the AutoValue
  * generated class. It is not guaranteed that an Extension's generated class will be the final
- * class in the inheritance hierarchy, unless its {@link #mustBeFinal(Context)} method returns true.
+ * class in the inheritance hierarchy, unless its
+ * {@link com.google.auto.value.extension.AutoValueExtension#mustBeFinal(Context)} method returns true.
```
eamonnmcmanus: I'm not seeing the purpose of these `{@link}` rewrites. The previous versions work for me. Is there a context where they don't?

eamonnmcmanus:
Sorry about the delay in getting to this. I've pulled it into our local repository but before committing it I want to restructure the code somewhat. The AutoValueProcessor.processType method in particular is getting a bit unwieldy. I'd also like to get rid of the warning for abstract methods like `writeToParcel(Parcel, int)`, which implies generalizing the support for consuming properties so that you can also consume other abstract methods.

Rainer-Lang:
Would be good to have this merged.

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
rharter: I would expect that would pass since you're suppressing all processor warnings.  The point of this test is to ensure that those two expected warnings, **and no others**, are thrown.  In the error case, a warning is thrown for every property, but `.withCompilerOptions("-Xlint:-processing")` would suppress those.

value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
rharter: Yeah, I'm not sure why those got in there.  Works fine for me, too.  I've reverted them.

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
tbroyer: Er, no, `-Xlint:-processing` is only about the warning wrt annotations that haven't been _claimed_ by a processor (and most processors, including Auto ones, don't claim annotations at all).
https://docs.oracle.com/javase/7/docs/technotes/tools/solaris/javac.html#xlintwarnings

eamonnmcmanus:
I'm going to close this PR because our processes mean that we incorporate changes internally at Google rather than on github. I've made a chunk of changes to extensions, partly based on this PR, which will appear at our next push.

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
rharter: ahh, yes, good point.  My misunderstanding.

PR #311, opened 2016-03-21
Rely on Guava 19.0 and use CharMatcher.whitespace()

cgruber:
CharMatcher.WHITESPACE will be removed in a future version of Guava (improves performance on android - not relevant to auto-value, but still nicer).

-------------
MOE_MIGRATED_REVID=116549017


eamonnmcmanus:
LGTM

PR #312, opened 2016-03-21
Add an explicit check for @AutoValue class being private

cgruber:
Add an explicit check for @AutoValue class being private, since otherwise that leads to a puzzling experience in IDEs.

-------------
MOE_MIGRATED_REVID=116771311


eamonnmcmanus:
LGTM

PR #313, opened 2016-03-21
Improve duplicate descriptor checking. 

cgruber:
When checking FactoryMethodDescriptor's and ImplementationMethodDescriptor's for duplicates, remove the ImplementationMethodDescriptor and set the FactoryMethodDescriptor to be overriding and have the same visibility.

  - The test case, which is now modified to make the concrete class package-private, would not have worked before, because the create methods were given package-private visibility instead of public visibility.
  - This also makes it so that these duplicate methods have the @Override annotation.

-------------
MOE_MIGRATED_REVID=117186640


eamonnmcmanus:
LGTM

Issue #200, opened 2015-03-04
AutoValue: Crash with seemingly normal code


kenzierocks:
I made a small change in some code recently and it caused the crash below. The subset of the project that caused the crash is available on https://github.com/TechShroom/Slithering-Latte/tree/showbug/av-crash. The specific change is https://github.com/TechShroom/Slithering-Latte/commit/c3dbdc1b13956314860b22e272067378efb67feb

```
@AutoValue processor threw an exception: java.lang.IllegalArgumentException
	at autovalue.shaded.org.objectweb.asm.ClassReader.<init>(Unknown Source)
	at autovalue.shaded.org.objectweb.asm.ClassReader.<init>(Unknown Source)
	at autovalue.shaded.org.objectweb.asm.ClassReader.<init>(Unknown Source)
	at com.google.auto.value.processor.AbstractMethodLister.abstractNoArgMethodsX(AbstractMethodLister.java:50)
	at com.google.auto.value.processor.AbstractMethodLister.abstractNoArgMethods(AbstractMethodLister.java:43)
	at com.google.auto.value.processor.EclipseHack$BinaryPropertyOrderer.determinePropertyOrder(EclipseHack.java:240)
	at com.google.auto.value.processor.EclipseHack.reorderProperties(EclipseHack.java:98)
	at com.google.auto.value.processor.EclipseHack.reorderProperties(EclipseHack.java:86)
	at com.google.auto.value.processor.AutoValueProcessor.defineVarsForType(AutoValueProcessor.java:424)
	at com.google.auto.value.processor.AutoValueProcessor.processType(AutoValueProcessor.java:369)
	at com.google.auto.value.processor.AutoValueProcessor.process(AutoValueProcessor.java:127)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.handleProcessor(RoundDispatcher.java:139)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.round(RoundDispatcher.java:110)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.BaseAnnotationProcessorManager.processAnnotations(BaseAnnotationProcessorManager.java:159)
	at org.eclipse.jdt.internal.apt.pluggable.core.dispatch.IdeAnnotationProcessorManager.processAnnotations(IdeAnnotationProcessorManager.java:134)
	at org.eclipse.jdt.internal.compiler.Compiler.processAnnotations(Compiler.java:851)
[...]

PR #314, opened 2016-03-21
Broaden exception catching in EclipseHack

cgruber:
In EclipseHack, catch any exception from propertyOrderer.determinePropertyOrder(), and not just IOException. This should be sufficient to work around an ASM crash reported in #200.

-------------
MOE_MIGRATED_REVID=117612741


eamonnmcmanus:
LGTM

PR #315, opened 2016-03-21
Bump the development version of auto-value

cgruber:


eamonnmcmanus:
LGTM

tbroyer:
FWIW there's a javapoet 1.6.1 rolling out with a fix to string escaping (regression from 1.5 IIUC)

ronshapiro:
@tbroyer the regression was in 1.6.0. I'm updating auto-factory, but I don't think anything's been synced yet, has it?

ronshapiro:
Ah, it looks like the commit list has changed in the interim. @cgruber do you know what happened? Either way I'll be updating to javapoet 1.6.1 once the tests pass Christian approves it :)

PR #317, opened 2016-04-05
Moe writing branch from 9084efc327fad709db7e72f9fa20c5feacde395a

cgruber:
A few changes:

  * Updates information w.r.t . Auto-Value's 1.2 release, adds a changelog (with only 1.2 specific info for now), and cleans up some other things.
  * ImmutableMapBuilder is removed due to a general guava cleanup
  * Javapoet is upgraded (for AutoFactory) but then rolled back.  It's import needed more massaging of
    internal users before we could safely keep it at 1.6.0, so that change will be forthcoming.


value/CHANGES.md
```diff
@@ -0,0 +1,59 @@
...
+## 1.1 → 1.2
+
+### Functional changes
+
+  * A **provisional** extension API has been introduced. This **will change**
+    in a later release. If you want to use it regardless, see the
+    [AutoValueExtension] class.
+
+  * Properties of primitive array type (e.g. `byte[]`) are no longer cloned
+    when read. If your @AutoValue class includes an array property, by default
```
eamonnmcmanus: `@AutoValue` should be quoted. But we can fix that later.

eamonnmcmanus:
LGTM

PR #322, opened 2016-04-13
Removing inaccurate JavaDocs reference

lucastsa:
AutoValueExtension's `generateClass` JavaDocs referred to an inexistent `classToImplement`.


eamonnmcmanus:
Thanks! We'll make that change internally and it will be in the next push to github.

Issue #278, opened 2015-10-21
AutoValue: Support Optional<T> similar to @Nullable


mfulton26:
I'd like to have "optional" value properties but instead of annotating my properties with `@Nullable` and then checking for `null` when using the property I'd rather use [`java.util.Optional<T>`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) and/or [`com.google.common.base.Optional<T>`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Optional.html).

Here's what I'm doing today:
```java
import com.google.auto.value.AutoValue;

import java.util.Optional;

class Example {
  @AutoValue
  abstract static class Animal {
    static Builder builder() {
      return new AutoValue_Example_Animal.Builder()
          .name(Optional.empty)
          .numberOfLegs(Optional.empty());
    }

    abstract Optional<String> name();
    abstract Optional<Integer> numberOfLegs();

    @AutoValue.Builder
    abstract static class Builder {
      Builder name(String s) { return name(Optional.ofNullable(s)); }
      abstract Builder name(Optional<String> s);
      Builder name(Integer n) { return name(Optional.ofNullable(n)); }
      abstract Builder numberOfLegs(Optional<Integer> n);
      abstract Animal build();
    }
  }
}
```

Here's what I want to do:
```java
import com.google.auto.value.AutoValue;

import java.util.Optional;

class Example {
  @AutoValue
  abstract static class Animal {
    static Builder builder() {
      return new AutoValue_Example_Animal.Builder();
    }

    abstract Optional<String> name();
    abstract Optional<Integer> numberOfLegs();

    @AutoValue.Builder
    abstract static class Builder {
      abstract Builder name(String s);
      abstract Builder numberOfLegs(Integer n);
      abstract Animal build();
    }
  }
}
```

[...]

PR #323, opened 2016-04-14
Removes toBuilder method from Extension Context properties method

lucastsa:
Current implementation returns `toBuilder` as a property in the `AutoValueExtension.Context` `properties()` passed to extensions, which causes problems when using extensions on classes that use Builders.

This pull requests removes the toBuilder method from the properties in the same fashion that was already being done in the `defineVarsForType` method.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
+++ b/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
@@ -362,8 +362,13 @@ public class AutoValueProcessor extends AbstractProcessor {
     ImmutableSet<ExecutableElement> methods =
         getLocalAndInheritedMethods(type, processingEnv.getElementUtils());
     ImmutableSet<ExecutableElement> methodsToImplement = methodsToImplement(type, methods);
-    ImmutableBiMap<String, ExecutableElement> properties =
-        propertyNameToMethodMap(methodsToImplement);
+
+    BuilderSpec builderSpec = new BuilderSpec(type, processingEnv, errorReporter);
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
Thanks for this contribution! However, I have a restructuring of the Extensions support that is close to completion and it will solve this problem among others.

lucastsa:
Thanks for letting me know, @eamonnmcmanus. Does the restructuring include information about the class builder to the extension?

eamonnmcmanus:
Not currently. Do you have a particular need that might be addressed by that?

lucastsa:
Possibly. I've seen the exchanges in #278 and realized that it is unlikely that a solution where `setX(T)` is going to be implemented, as the consensus seems to be that setX(null) should not mean `Optional.empty()` / `Optional.absent()`. An alternative solution to who uses that pattern `Optional<T> getX()` + `Builder.setX(T)`, where the non-serializable `java.util.Optional` is only used as a return type, could be provided by an extension, that would require implementing a setter in a builder subclass and possibly "consume" those properties so that AutoValue doesn't complain about property type mismatch.

eamonnmcmanus:
I haven't given the matter much thought, but since the generated Builder is nested inside the generated AutoValue_Foo, and extensions work by subclassing the former, they could potentially subclass the latter too. Currently the generated Builder class is final but we could turn that off when extensions are present. I think we'd need to have some clear use cases first. I'm not sure that "knowing whether properties have been set" is a good enough one, because I think the planned `Optional<T>` solution for it is going to be enough for most users.

Issue #331, opened 2016-05-25
AutoValue.java RentionPolicy(RetentionPolicy.SOURCE)


williamwebb:
If there is no downside to increasing the retention level it would be extremely helpful to have this annotation retained at the Runtime for Deserialization purposes.

As shown [here](https://gist.github.com/JakeWharton/0d67d01badcee0ae7bc9), currently to achieve deserialization an additional annotation is required. Which while this works requires users to know to use this annotation. If AutoValue.java were retained at runtime there would be a concrete relationship that could be used and enforced by the AutoValue annotation processor.

eamonnmcmanus:
I'm not sure this is sufficiently strong motivation. The retention is SOURCE because this is an implementation detail. Clients of an AutoValue class should neither know nor care that it is one.

Assigning to @kevinb9n for a second opinion.

JakeWharton:
There's also https://github.com/rharter/auto-value-gson/ now which uses the built-in extension mechanism to achieve the same effect as my Gist without reflection.

williamwebb:
@JakeWharton only reasoning for not going with that option is it again requires the uses of AutoValue to know to add the static method to return the TypeAdapter. For something that would ideally be transparent.

kevinb9n:
I think that extension or another like it is the right place to work this out.

Éamonn's statement, "Clients of an AutoValue class should neither know nor care that it is one", could be fairly described as THE founding principle of AutoValue. The thing that made us realize our idea was different from all the rest. The annotation is not even @Documented, let alone visible to non-source-level tools.

williamwebb:
👍

PR #332, opened 2016-05-25
Increase retention level of the AutoValue annotation to facilitate deserialization

williamwebb:
Facilitates the ability to deserialize AutoValue objects at runtime.

For issue #331


williamwebb:
I signed it!

PR #338, opened 2016-05-31
Update 2016 05 31

### LINKED ISSUE #337: Unnecessary semicolon in generated AutoAnnotation_*#appendEscaped
Our project has the following @AutoAnnotation method:
https://gerrit.googlesource.com/gerrit/+/136aab1f0abea5675af7fcac4eac2e90f15e61fa/gerrit-sshd/src/main/java/com/google/gerrit/sshd/Commands.java#50

Building this with auto-value 1.2 produces the following appendEscaped method:

```java
  private static void appendEscaped(StringBuilder sb, char c) {
    switch (c) {
    case '\\':
    case '"':
    case '\'':
      sb.append('\\').append(c);
      break;
    case '\n':
      sb.append("\\n");
      break;
    case '\r':
      sb.append("\\r");
      break;
    case '\t':
      sb.append("\\t");
      break;
    default:
      if (c < 0x20) {
        sb.append('\\');
        appendWithZeroPadding(sb, Integer.toOctalString(c), 3);;
      } else if (c < 0x7f || Character.isLetter(c)) {
        sb.append(c);
      } else {
        sb.append("\\u");
        appendWithZeroPadding(sb, Integer.toHexString(c), 4);
      }
      break;
    }
  }
```

Note the extra semicolon in the `if (c < 0x20)` block.

(If you want to reproduce this locally by building the whole Gerrit project, you'll need to follow the [Eclipse project setup instructions](https://gerrit-review.googlesource.com/Documentation/dev-eclipse.html), then `cat .apt_generated/com/google/gerrit/sshd/AutoAnnotation_Commands_named.java`. Unfortunately, while the non-IDE Buck-driven build is easier to set up, it doesn't produce the generated source code as an output.)

eamonnmcmanus:
# AutoValue, AutoAnnotation

* Make AutoValueProcessor's constructor that takes extensions public so that it can be used by compile-testing tests of AutoValueExtensions.
* Rework the javadoc of AutoValue.java so that it doesn't get stripped during export.
* Remove a redundant semicolon in the @AutoAnnotation template, and introduce a blank line because indentation disappears on the line after a Velocity comment. Fixes #337.
* Make copyright notices consistent.

# AutoFactory

* Add a note that states that you have to use @Qualifier instead of @BindingAnnotation in order for the generated code to work correctly.

# AutoCommon

* Add text to the specification of MoreElements.getLocalAndInheritedMethods to point out that the returned set may contain more than one method with the same signature.

### DIFF
--- a/common/src/main/java/com/google/auto/common/MoreElements.java
+++ b/common/src/main/java/com/google/auto/common/MoreElements.java
@@ -245,6 +245,12 @@ public final class MoreElements {
    * result. So if {@code type} defines {@code public String toString()}, the returned set will
    * contain that method, but not the {@code toString()} method defined by {@code Object}.
    *
+   * <p>The returned set may contain more than one method with the same signature, if
+   * {@code type} inherits those methods from different ancestors. For example, if it
+   * inherits from unrelated interfaces {@code One} and {@code Two} which each define
+   * {@code void foo();}, and if it does not itself override the {@code foo()} method,
+   * then both {@code One.foo()} and {@code Two.foo()} will be in the returned set.
+   *
    * @param type the type whose own and inherited methods are to be returned
    * @param elementUtils an {@link Elements} object, typically returned by
    *     {@link javax.annotation.processing.AbstractProcessor#processingEnv processingEnv}<!--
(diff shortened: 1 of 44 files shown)


cgruber:
LGTM

PR #339, opened 2016-06-04
Only use one StringBuilder per Builder.build call

kenzierocks:
Rewrites the builder to use a single `StringBuilder` when building the `missing` variable. I'm not sure if this would have any noticeable performance impact (positive or negative), but I'd thought I'd PR it anyways since it has so little changes.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/autovalue.vm
+++ b/value/src/main/java/com/google/auto/value/processor/autovalue.vm
@@ -329,17 +329,17 @@ ${gwtCompatibleAnnotation}
 
   #if (!$builderRequiredProperties.empty)
 
-      String missing = "";
+      StringBuilder missing = new StringBuilder();
 
     #foreach ($p in $builderRequiredProperties)
 
       if ($p == null) {
-        missing += " $p.name";
+        missing.append(" $p.name");
       }
 
     #end
 
-      if (!missing.isEmpty()) {
+      if (missing.length() > 0) {
         throw new IllegalStateException("Missing required properties:" + missing);
       }
   #end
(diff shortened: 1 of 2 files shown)


eamonnmcmanus:
Yeah, I'm not really convinced. You're saving a little time in the exception case, but sacrificing some time in the non-exception case, since you're constructing a StringBuilder() that will be discarded. I don't think that's a good trade.

kenzierocks:
Closing as I don't think this is really worth anything.

blipinsk:
@eamonnmcmanus could you elaborate a bit more on why this wasn't used? I'm curious. The `missing` String is also discarded and if the StringBuilder was initialized with the proper initial capacity (we can calculate the upper boundary as a sum of lengths of variable names plus the additional spaces) for me it seems like a good idea.

```
@Override 
public Movie build() {
    StringBuilder missing = new StringBuilder(32);
    if (title == null) {
      missing.append(" title");
    }
    if (director == null) {
      missing.append(" director");
    }
    if (cast == null) {
      missing.append(" cast");
    }
    if (releaseDate == null) {
      missing.append(" releaseDate");
    }
    if (missing.length() > 0) {
      throw new IllegalStateException("Missing required properties:" + missing.toString());
    }
    return new AutoValue_Movie(this.title, this.director, this.cast, this.releaseDate);
  }
```

JakeWharton:
That makes the cost even worse in the non-exception case.

blipinsk:
Yea, right. Assuming the usage of the builder is correct there won't be any string additions what so ever. And if someone is not setting all parameters or is setting `null` values, the string appending inefficiency is not really important because it's ending with an `IllegalStateException`. Thanks for the tip @JakeWharton

PR #346, opened 2016-06-15
Update 2016-06-15

### LINKED ISSUE #345: Ctor null checks omitted when builder is present, but ctor still package-visible.
Given a model
```java
@AutoValue
public abstract class Whatever {
  public abstract String whatever();

  @AutoValue.Builder
  interface Builder {
    Builder whatever(String whatever);
    Whatever build();
  }
}
```
AutoValue 1.3-rc1 will generate:
```java
@Generated("com.google.auto.value.processor.AutoValueProcessor")
 abstract class $AutoValue_Whatever extends Whatever {

  private final String whatever;

  $AutoValue_Whatever(
      String whatever) {
    this.whatever = whatever;
  }

  // omitted

  static final class Builder implements Whatever.Builder {
    private String whatever;
    Builder() {
    }
    Builder(Whatever source) {
      this.whatever = source.whatever();
    }
    @Override
    public Whatever.Builder whatever(String whatever) {
      this.whatever = whatever;
      return this;
    }
    @Override
    public Whatever build() {
      String missing = "";
      if (whatever == null) {
        missing += " whatever";
      }
      if (!missing.isEmpty()) {
        throw new IllegalStateException("Missing required properties:" + missing);
      }
      return new AutoValue_Whatever(
          this.whatever);
    }
  }

}
```
As you can see the constructor is still package scoped which the application code could use to insert a `null` value for `whatever`. This constructor is also used by extensions which are not null-checking their values since they rely on the constructor to validate the nullability constraints.

eamonnmcmanus:
* The constructor of an @AutoValue class that has a builder needs to have null checks if there are extensions, because in that case the constructor is package-private rather than private. Fixes #345.

* When scanning a Builder for referenced types, don't include non-abstract methods. The referenced types are used to determine the imports that will be needed by the implementing class, and that class never references non-abstract methods.


cgruber:
Might want to reword my commit, since the only public piece is the removal of the nav.

cgruber:
Or if you don't care enough LGTM.  It's just inaccurate. /shruggie.

eamonnmcmanus:
Yeah, I don't think anybody cares.

Issue #355, opened 2016-07-10
auto-value Error: Nullable is not a repeatable annotation type


natros:
This code is falling with Error:(29, 3) java: org.jetbrains.annotations.Nullable is not a repeatable annotation type

``` java
import com.google.auto.value.AutoValue;
import org.jetbrains.annotations.Nullable;

@AutoValue
public abstract class User {
    abstract Integer userId();
    @Nullable abstract String login();
    abstract String name();
    public static Builder builder() {
        return new AutoValue_User.Builder();
    }

    @AutoValue.Builder
    public abstract static class Builder {
        abstract Builder userId(Integer userId);
        abstract Builder login(String login);
        abstract Builder name(String name);
        abstract User build();
    }
}
```

And the generated code

``` java
import javax.annotation.Generated;
import org.jetbrains.annotations.Nullable;

@Generated("com.google.auto.value.processor.AutoValueProcessor")
 final class AutoValue_User extends User {

  private final Integer userId;
  private final String login;
  private final String name;

  private AutoValue_User(
      Integer userId,
      @Nullable String login,
      String name) {
    this.userId = userId;
    this.login = login;
    this.name = name;
  }

  @Override
  Integer userId() {
    return userId;
  }

  @Nullable
  @Nullable
  @Override
  String login() {
    return login;
  }

  @Override
  String name() {
    return name;
  }

  @Override
  public String toString() {
    return "User{"
        + "userId=" + userId + ", "
        + "login=" + login + ", "
        + "name=" + name
        + "}";
  }

  @Override
  public boolean equals(Object o) {
    if (o == this) {
      return true;
    }
    if (o instanceof User) {
      User that = (User) o;
      return (this.userId.equals(that.userId()))
           && ((this.login == null) ? (that.login() == null) : this.login.equals(that.login()))
           && (this.name.equals(that.name()));
    }
    return false;
  }

  @Override
  public int hashCode() {
    int h = 1;
    h *= 1000003;
    h ^= this.userId.hashCode();
    h *= 1000003;
    h ^= (login == null) ? 0 : this.login.hashCode();
    h *= 1000003;
    h ^= this.name.hashCode();
    return h;
  }

  static final class Builder extends User.Builder {
    private Integer userId;
    private String login;
    private String name;
    Builder() {
    }
    Builder(User source) {
      this.userId = source.userId();
      this.login = source.login();
      this.name = source.name();
    }
    @Override
    public User.Builder userId(Integer userId) {
      this.userId = userId;
[...]

PR #356, opened 2016-07-13
Fix for issue #355 "auto-value Error: Nullable is not a repeatable annotation type"

brychcy:


value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -216,6 +216,7 @@ private String generatedSubclassName(TypeElement type, int depth) {
 
     private ImmutableList<String> buildAnnotations(TypeSimplifier typeSimplifier) {
       ImmutableList.Builder<String> builder = ImmutableList.builder();
+      Set<Name> methodAnnotations = null;
```
tbroyer: How about storing the `AnnotationMirror`s directly? According to [the javadoc](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/AnnotationMirror.html), it supports `equals()`

brychcy: It would probably work, but unfortunately its javadoc mentions only equals, not hashCode. For Name, the [javadoc](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/Name.html) does mention hashCode, too

eamonnmcmanus: I think it is fair to assume that any public JDK class or interface that implements `equals(Object)` also implements a corresponding `hashCode()`. Otherwise it would be violating the contract of `hashCode()`.

brychcy: I just tried it, it doesn't work at all, at least for javac. The same (source-code) annotation as method annotionMirror is not even equal() to the annotationMirror returned for the type.

eamonnmcmanus:
Issue #355 was fixed three years ago. Thanks for this contribution, which was helpful to the fix.

PR #351, opened 2016-06-25
fixed typo in code example

ersin-ertan:
added " to complete the string

PR #371, opened 2016-08-30
Changes from Google's internal version, up to 2016-08-29.

### LINKED ISSUE #187: Propagate @GwtIncompatible to @AutoValue types, as is done for @GwtCompatible
Motivation: [Truth defines an `AutoValue` type that doesn't work under GWT.](https://github.com/google/truth/blob/3f974942eecfb0318740c0bce8bd50395a9db001/core/src/main/java/com/google/common/truth/Expect.java#L52) But the AutoValue processor still runs, and it still generates the implementation. The implementation refers to the `@GwtIncompatible` superclass, and we get compile errors:

    [ERROR] Line 46: Expect.ExpectationFailure cannot be resolved to a type
    [ERROR] Line 23: The method message() of type AutoValue_Expect_ExpectationFailure must override or implement a supertype method
    [ERROR] Line 29: The method cause() of type AutoValue_Expect_ExpectationFailure must override or implement a supertype method
    [ERROR] Line 47: Expect.ExpectationFailure cannot be resolved to a type
    [ERROR] Line 7: Expect.ExpectationFailure cannot be resolved to a type

(I haven't managed to reproduce this with Truth's own GWT tests. I'm currently seeing it only when I modify our Guava build, which currently does its own `@GwtIncompatible` stripping, to rely on the GWT compiler's new automatic stripping. [Update: Truth's tests do catch it after all. I just hadn't run them *after* switching to the GWT compiler's stripping.])

I propose that AutoValue copy the `@GwtIncompatible` annotation just as it does the `@GwtCompatible` annotation, with one small twist: We should also copy the annotation from enclosing types, since the GWT compiler treats it as applying to enclosed types.

I'm working on an internal CL now, so if all this sounds reasonable, I'll send it for review when it's ready.

### LINKED ISSUE #296: AutoFactory omits Provider in constructor call
The following code generates a factory which does not compile due to a wrong parameter in the constructor call.

```java
import javax.inject.Provider;
import com.google.auto.factory.Provided;

@AutoFactory
public class Root {
    public Root(Provider<Child> child, @Provided Injected injected) {
    }
}
```

The create method in the generated factory expects a parameter of type `Child` instead of `Provider<Child>` and calls the constructor with the wrong parameter.

```java
import javax.annotation.Generated;
import javax.inject.Inject;
import javax.inject.Provider;
@Generated("com.google.auto.factory.processor.AutoFactoryProcessor")
public final class RootFactory {
  private final Provider<Injected> injectedProvider;
  @Inject
  public RootFactory(Provider<Injected> injectedProvider) {
    this.injectedProvider = injectedProvider;
  }
  public Root create(Child child) {
    return new Root(child, injectedProvider.get());
  }
}
```

This bug seems to be only triggered when using an instance of javax.inject.Provider.

### LINKED ISSUE #326: AutoFactory does not handle varargs parameters
I'm trying to convert an existing type to be partially injected using autofactory, but our existing constructors and client code makes use of varargs to pass in differing numbers of arguments.

For example:
```java
@AutoFactory
class MyClass {
    public MyClass(@Provided InjectedParameter ip, RepeatedParameter... rp) {
        // ...
    }
}
```

However the signature of the corresponding `create` method is: `MyClassFactory.create(RepeatedParameter[] rp)` which rather defeats the point.

It's possible to work around this by changing the signature to accept a `List<RepeatedParameter>` and using `Arrays.asList()`, but it would be better if AutoFactory just supported varargs parameters directly.

eamonnmcmanus:
* Change the AutoValue version from 1.3-SNAPSHOT to 1.4-SNAPSHOT now that 1.3 has been released. Also change AutoValue's dependency on AutoCommon to 0.6 rather than 1.0-SNAPSHOT. If AutoValue later needs to pick up changes from a more recent AutoCommon, we should make an AutoCommon 0.7 at that point.

* Fix compile-time constant expression overflows in AutoAnnotation.

* Update the integration test pom.xml files to use a version of maven-jar-plugin that isn't broken on ubuntu, causing multi-minute jar times.  And while here, bump them all to 2.6 (2.5+ is fine, but may as well, and may as well be consistent.)

* Use named classes in MoreElements and MoreTypes to make clearer stack traces.

* Allow varargs on `@AutoFactory` methods. Fixes #326.

* Rename FactoryImplementingGenericInterfaceExtensionFactory file to match the name of the class.

* Make AutoValue.Builder examples use "set" prefixes, which makes them easier for readers to understand, and I believe is generally more common in the wild for builders, EJ notwithstanding.

* Fix Java source files so that they follow the new simplified import rules.

* Disambiguate provider field's names in generated factories.

* Avoid propagating an exception to the compiler if we fail to load AutoValue extensions due to a corrupt jar file in the class path.

* Allow `Provider<>` parameters on a `create()` method. Also check nullness of `Provider<>`s passed to `@Inject` constructors of an `@AutoFactory`. Fixes #296.

* Remove workarounds for bugs/missing features before JavaPoet 1.7.

* AutoCommon: Add a hook for processing over.

* Add `@AutoValue.CopyAnnotations`, which causes annotations to be copied from the annotated class to the generated class. Fixes #187.

* Fix a bug in AutoAnnotation when using Collections to pass values to generic array parameters. Without this AutoAnnotation would generate code that constructed arrays of generic types which would result in errors from javac like: `Cannot create a generic array of Class<? extends Annotation>`. Also, change the generated code to not presize the array when calling Collection.toArray, for hotspot vms this is likely a pessimization per http://shipilev.net/blog/2016/arrays-wisdom-ancients/ and I doubt `@AutoAnnotation` is used much outside of hotspot vms.

* Support `@Nullable` parameters for `@AutoFactory`.

* Fix typo noted by @ersin-ertan in https://github.com/google/auto/pull/351.


cgruber:
There are some commit descriptions that no longer make sense, but I think they're minor.  I'd fix'em, but I won't make you. :)

LGTM

PR #376, opened 2016-09-08
Templates are parsed using Template.parseFrom, not Template.from

cgruber:
(as noted by austinchen) 

-------------
MOE_MIGRATED_REVID=132505424


cgruber:
One of the three JVMs is failing with extra warnings. Not sure why. :/

eamonnmcmanus:
The failure looks like this:
```
testDoesntRaiseWarningForConsumedProperties(com.google.auto.value.processor.ExtensionTest)  Time elapsed: 0.156 sec  <<< FAILURE!
java.lang.AssertionError: Expected 0 warnings, but found the following 1 warnings:
Implicitly compiled files were not subject to annotation processing.
```

I suspect that's because the system in question has a newer JDK that includes this warning. I will look into what might be involved in suppressing the warning. Meanwhile, it clearly doesn't have anything to do with this PR, which is purely a documentation change.

PR #378, opened 2016-09-12
Support constructors that throw checked exceptions (Fixes #90).

### LINKED ISSUE #90: AutoFactory does not support constructors that throw checked exceptions.
Checked exceptions declared by the constructor should be added to the throw list for the generated create() method.

janstrohbeck:
This merge request adds checked exceptions that a constructor may throw to the throw list of the generated create() method.
The added code could probably be simpilfied more.


janstrohbeck:
I signed it!

Issue #365, opened 2016-08-02
Sporadic IOExceptions during parsedTemplateForResource


hzsweers:
We're seeing sporadic failures during builds from AutoValue during `parsedTemplateForResource`. We're really at a loss for what's going on, and are happy to dig in to anything you want us to try. Using AutoValue 1.3-rc2, one internal custom extension being used in Android projects. They look like this (both with gradle and buck builds):

```
Caused by: java.lang.AssertionError: java.io.IOException: Stream closed

	at com.google.auto.value.processor.TemplateVars.parsedTemplateForResource(TemplateVars.java:111)
	at com.google.auto.value.processor.AutoValueTemplateVars.<clinit>(AutoValueTemplateVars.java:184)
	at com.google.auto.value.processor.AutoValueProcessor.processType(AutoValueProcessor.java:441)
	at com.google.auto.value.processor.AutoValueProcessor.process(AutoValueProcessor.java:150)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:794)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:705)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.access$1800(JavacProcessingEnvironment.java:91)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1035)
	at com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1176)
	at com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1170)
	at com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:856)
	at com.sun.tools.javac.main.Main.compile(Main.java:523)
	... 34 more
Caused by: java.io.IOException: Stream closed

	at com.google.auto.value.processor.escapevelocity.Parser.next(Parser.java:141)
	at com.google.auto.value.processor.escapevelocity.Parser.<init>(Parser.java:84)
	at com.google.auto.value.processor.escapevelocity.Template.parseFrom(Template.java:57)
	at com.google.auto.value.processor.TemplateVars.parsedTemplateForResource(TemplateVars.java:107)
	... 45 more
```

```
java.lang.RuntimeException: java.lang.AssertionError: java.io.IOException: Stream closed

[...]

PR #386, opened 2016-09-29
Remove IO reads portion of velocity usage to avoid JDK bug exposed in #365

hzsweers:
This avoids the IO issue we see in https://github.com/google/auto/issues/365 by not parsing VM templates from local resource files.

Wanted to put this PR up for discussion, as I'm not sure if there's any strong feelings one way or the other around using resources vs. String constants. It's not quite as easy to edit on the fly but still fine if you use something like IntelliJ's language injection + code fragment editor.


eamonnmcmanus:
It does make me a bit sad not to have .vm files, which are supported in pretty much every editing environment. I wonder if we could make an annotation processor that just translates text files into string constants in Java source files. The only thing holding me back there is that it does complicate the build process for AutoValue somewhat.

eamonnmcmanus:
I think there might be a workaround for the existing approach. I've commented on #365.

hzsweers:
Closing in favor of #390 and its followup

PR #392, opened 2016-10-12
Sync out google internal changes.

cgruber:
Changes include:

  * Javadoc fixes
  * Update to ASM-5 for Java8 compatibility (should still work in Java7)
  * Support @LazyInit from error-prone in cases where @Memoized is too eager
  * Make AutoValue only import top-level types (and use relative references to nested types)
  * Switch to using the FormattingFiler from google-java-format.


eamonnmcmanus:
LGTM

PR #393, opened 2016-10-12
Latest changes from Google.

eamonnmcmanus:
* [common] Fix a number of incompatibilities between javac and ecj in the area of overrides, by reworking the logic of the explicit override implementation.

* [value] Add a workaround to Template.parsedTemplateForResource that triggers fall-back logic if we hit a JDK bug that causes spurious IOExceptions when reading resources from jars. Also, rewrite TemplateVarsTest to use JUnit4 and Truth. Fixes https://github.com/google/auto/issues/365.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -75,6 +75,12 @@
       <version>${truth.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jdt.core.compiler</groupId>
+      <artifactId>ecj</artifactId>
+      <version>4.5.1</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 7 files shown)


cgruber:
LGTM

hzsweers:
Awesome, thanks so much! Do you know if you're planning a new release any time soon, or should we fork this internally for now?

cgruber:
Which sub-project?  We don't really have a heart-beat release schedule as Dagger now does, but we can release on-demand, really.

hzsweers:
Just autovalue, we've currently had to force all our models to the same module to avoid the bug this resolves

eamonnmcmanus:
@cgruber, a release would make sense now, but we'd have to release auto-common first because we're depending on an API change there.

cgruber:
That's fine.  I can do so tomorrow.

cgruber:
Auto-Common 0.8 and AutoValue 1.4 RC1 have been released (the latter just a
few minutes ago, so give it an hour to sync to sonatype's servers).  If
there are no issues with 1.4-rc1 for a few days I'll re-release as 1.4.

hzsweers:
perfect, will update our projects today

PR #394, opened 2016-10-20
Add builder setter type to error message when type of parameter is not equivalent to target type

gildor:


gildor:
Parameter type information in error message is especially useful when you use code generated by other tools. For exaple in my case I used Kotlin for AutoValue builders and couldnot figure out error without additional logging.

eamonnmcmanus:
Thanks for the PR! I'm applying this change manually to Google's internal version and it should be present in a forthcoming sync from that version.

Issue #398, opened 2016-10-26
@Memoized expects javax.annotation.Generated to be present


blipinsk:
The new `MemoizeExtension` expects `javax.annotation.Generated` to be present in the project. 
If you don't have it, build fails.

![generated_missing](https://cloud.githubusercontent.com/assets/6371716/19720486/9b6ce75a-9b6e-11e6-8851-561f7b47ae70.png)

Need a PR for that?

blipinsk:
Since those are just like few lines of code, here's a pull request 😉

commonsguy:
Note that his is particularly relevant for Android, which does not have `javax.annotation.Generated`, meaning that `@Memoize` appears to be unusable on Android.

eamonnmcmanus:
This is a duplicate of #503, which is fixed in AutoValue 1.5.1.

PR #399, opened 2016-10-26
Adding @Generated annotation only when it's present on the classpath

blipinsk:
The new `MemoizeExtension` wasn't taking into account the possibility that `@Generated` annotation isn't present on classpath (see Issue #398).

This pull request fixes the problem (using the same convention that was previously used for `com.google.errorprone.annotations.concurrent.LazyInit`).


blipinsk:
I signed it!

blipinsk:
Not sure why travis fails... Those tests pass for me locally. More so the `ExtensionTest` fails, which shouldn't be affected by my changes to MemoizeExtension.java.

eamonnmcmanus:
Looks like we fixed this independently in AutoValue 1.5.1. :-/

blipinsk:
Glad it was fixed 😋

PR #407, opened 2016-12-05
Moe sync 12/05/2016

ronshapiro:


eamonnmcmanus:
I think we need to figure out why ExtensionTest is failing with this error: Implicitly compiled files were not subject to annotation processing. But only on Java 7, not on Java 6 nor on Java 8.

ronshapiro:
It's failing in openjdk7, but not oraclejdk7 or oraclejdk8... it's bizarre because it's only compiling that one file, and just in that one test.

Do you remember why you added `-Xlint:-processing` to those three tests? those seem to be the only ones that it's failing for.

Separately, do we care about compiling with jdk7 anymore? Everyone should be using javac8 even if they pass `-source 7 -target 7`

eamonnmcmanus:
The `-Xlint:-processing` turns off warnings about annotations that were not consumed by any annotation processor.

We can certainly discuss no longer compiling with jdk7. Do you want to open an issue for that? I'd like to fix the issue at hand if we can, though.

tbroyer:
> Everyone should be using javac8 even if they pass `-source 7 -target 7`.

Nobody should ever _“pass `-source` and `-target`”_ to javac unless they also use `-bootclasspath`; and actually javac8 warns you about it (afaik, javac9's `-release` should finally save us all from having many JDKs installed locally)
This is not only hypothetical/theoretical, it can really break at runtime: see http://developer-blog.cloudbees.com/2014/12/beware-siren-target-call.html and http://www.draconianoverlord.com/2014/04/01/jdk-compatibility.html

> Do you remember why you added `-Xlint:-processing` to those three tests? those seem to be the only ones that it's failing for.

Those tests are checking for `compileWithoutWarning` (they're the only 3 tests doing so), so without `-Xlint:-processing` they'd have to check for the warning about annotations that haven't been claimed instead.

I wonder if those errors aren't false positives / flakes. It'd be interesting to know which class / source file was implicitly compiled. I actually wonder if compile-testing shouldn't default to using an empty _source path_ to prevent those kind of implicit compilation to begin with.

eamonnmcmanus:
My inclination is to add "-implicit:class" to the tests, which is supposed to silence that particular warning. I agree that it would be interesting to know what the implicitly-compiled file in question is, but I'm not able to reproduce the failure on my own machine so I don't see a straightforward way to find out.

tbroyer:
FWIW, I couldn't reproduce either [1], so it might be related to the version of Maven being used, or the way the build is done on Travis (building once without tests, then rebuilding with the tests but without cleaning; in my projects, I disabled the first run and just live with the `Downloading:`/`Downloaded:` lines in the log).

[1] I used Docker to get an OpenJDK 7, as my Ubuntu only has OpenJDK 8:
```
docker run -it --rm -v "$PWD":/usr/src/mymaven -w /usr/src/mymaven maven:3-jdk-7 mvn -B -U -f build-pom.xml clean  verify --fail-at-end -Dsource.skip=true -Dmaven.javadoc.skip=true
```

ronshapiro:
@eamonnmcmanus LGTY?

eamonnmcmanus:
Great, thanks!

PR #410, opened 2016-12-13
Sync 2016-12-13: Finish the generalization of property builders

eamonnmcmanus:
* When property `Bar bar()` has a builder `BarBuilder barBuilder()`, don't require there to be a `Bar.toBuilder()` unless it is actually needed.

* Document the new "generalized property builders" feature for @AutoValue.

* Release the functional tests for AutoAnnotation, which were not being included in the open-source build because of a faulty configuration.


ronshapiro:
LGTM

PR #416, opened 2016-12-20
Moe sync

ronshapiro:


eamonnmcmanus:
Bummer. Compile Testing now requires Java 8. Perhaps we can exclude the tests that need it from running on earlier Java versions?

ronshapiro:
To confirm - there are still users which are using `javac7` and can't upgrade to `javac8`?

eamonnmcmanus:
I don't know. Maybe it is time to retire the testing on jdk7 and jdk6. However, in that case we should at least configure things so that tests on jdk8 use a -bootclasspath from jdk6.

eamonnmcmanus:
I think we can merge this in now, and fix up the jdk7 business later, probably by configuring things so that we only test on jdk8 but with an appropriate -bootclasspath for the compiler.

PR #417, opened 2016-12-21
Sync 2016 12 21

eamonnmcmanus:
* Rewrite TypeSimplifierTest to use compile-testing and get rid of its hairy custom annotation processor.

* Remove the exclusion of Truth from the dependencies of Compile Testing.

* Tweak AutoValue's Maven configuration so that it can work with Java 7 builds, by choosing an older version in that case and excluding tests that require the newest version.

### DIFF
--- a/value/pom.xml
+++ b/value/pom.xml
@@ -34,6 +34,8 @@
 
   <properties>
     <guava.version>19.0</guava.version>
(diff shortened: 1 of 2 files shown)


ronshapiro:
LGTM

PR #428, opened 2017-01-28
Update AutoValue "Latest version" to 1.3

peihe:


peihe:
R: @eamonnmcmanus

eamonnmcmanus:
Thanks! Our processes are such that we'll be making this change internally and pushing it out, but this PR will be referenced.

PR #431, opened 2017-02-08
Moe sync

ronshapiro:


eamonnmcmanus:
Crud, it looks as if ExtensionTest can't use the latest version of compile-testing because we're still building to be Java 6 compatible. I'll need to rework that.

PR #433, opened 2017-02-16
Moe sync

ronshapiro:


eamonnmcmanus:
I'm merging this even though it fails the Travis build. The failure is only for GWT and only on Java 7. We should figure out what's going on shortly.

tbroyer:
The problem is that GWT tests are run in DevMode. GWT 2.8  uses Java 8 sources, whichever the JVM, and DevMode will compile to Java 8 classes. In GWT 2.8, tests should run in production mode by default though. Try updating the gwt-maven-plugin to 1.0-rc-6.

eamonnmcmanus:
My proposed fix is to use Maven profiles to use GWT 2.7 with Java 7 and GWT 2.8 with Java 8. We do want to use GWT 2.8 where possible because the serialization test doesn't work properly with IPv6 on earlier versions. However, once we have released AutoValue 1.4, I'm also proposing that we drop Java 7 support everywhere (though we'll still support generating Java 6 compatible code).

tbroyer:
I think you'll want to run tests in prod mode anyway (devmode is deprecated, it's already gone in the Google-internal version of GWT), and running tests in prod mode will (should) fix the build with Java 7 so you won't need Maven profiles.

Updating the gwt-maven-plugin will automatically let GWT tests run with the default configuration (which means prod mode for GWT 2.8); they were forced to dev mode until 1.0-rc-4 (included).

So I'd propose simply updating gwt-maven-plugin and profit from the side effects to see the build fixed in Java 7.

eamonnmcmanus:
I don't really understand GWT well enough to follow what you are saying. Do you mean that we can set things up so that we can keep GWT 2.8 [here](https://github.com/google/auto/blob/master/value/src/it/gwtserializer/pom.xml) and yet have the tests run on Java 7? Does GWT 2.8 not require Java 8?

PR #434, opened 2017-02-22
Update gwtserializer/pom.xml to run tests in prod mode

tbroyer:
Running tests in prod mode is actually a side-effect of updating
the gwt-maven-plugin to 1.0-rc-6.

Also update guava to the same version as guava-gwt, remove
GWT version (redundant with dependencyManagement), and
update JUnit.

### DIFF
--- a/value/src/it/gwtserializer/pom.xml
+++ b/value/src/it/gwtserializer/pom.xml
@@ -118,7 +118,7 @@
       <plugin>
         <groupId>net.ltgt.gwt.maven</groupId>
         <artifactId>gwt-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


tbroyer:
+@eamonnmcmanus

eamonnmcmanus:
Sorry, I hadn't seen this before replying to your earlier comment.

eamonnmcmanus:
This is great! The passing Travis result shows that it fixes the problem, and it is plainly much better than what I was attempting to do with Maven profiles. We won't be merging this PR as such, but I'm copying it to Google's internal repo and it should appear (with due attribution) the next time we sync out. Thanks!

PR #439, opened 2017-03-01
Catch Inflater NPE in templatevars

hzsweers:
We started seeing the IO issues that #393 was intending to fix again, this time because there's apparently an NPE that gets thrown when the stream is closed ಠ_ಠ.

```
Caused by: java.lang.NullPointerException: Inflater has been closed
	at java.util.zip.Inflater.ensureOpen(Inflater.java:389)
	at java.util.zip.Inflater.inflate(Inflater.java:257)
	at java.util.zip.InflaterInputStream.read(InflaterInputStream.java:152)
	at java.io.FilterInputStream.read(FilterInputStream.java:133)
	at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
	at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
	at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
	at java.io.InputStreamReader.read(InputStreamReader.java:184)
	at java.io.BufferedReader.read1(BufferedReader.java:210)
	at java.io.BufferedReader.read(BufferedReader.java:286)
	at java.io.BufferedReader.fill(BufferedReader.java:161)
	at java.io.BufferedReader.read(BufferedReader.java:182)
	at java.io.LineNumberReader.read(LineNumberReader.java:126)
	at com.google.auto.value.processor.escapevelocity.Parser.next(Parser.java:140)
	at com.google.auto.value.processor.escapevelocity.Parser.<init>(Parser.java:83)
	at com.google.auto.value.processor.escapevelocity.Template.parseFrom(Template.java:56)
	at com.google.auto.value.processor.TemplateVars.templateFromInputStream(TemplateVars.java:138)
	at com.google.auto.value.processor.TemplateVars.parsedTemplateForResource(TemplateVars.java:112)
	at com.google.auto.value.processor.AutoValueTemplateVars.<clinit>(AutoValueTemplateVars.java:192)
```

PR #442, opened 2017-03-06
Sync 2016 03 06

eamonnmcmanus:
* Give a better error message when @AutoValue.Builder is applied to a non-static class.
* In the hack to work around spurious exceptions reading resources, handle NullPointerException in addition to IOException, since apparently it can happen too. This was reported in https://github.com/google/auto/pull/439.
* Avoid using keywords as parameter names for code generated by MemoizeExtension
* Fix a bug in the code for AutoValue extensions that don't generate a class in the AutoValue hierarchy. If there are extensions but none of them generates code then our AutoValue implementation class will be abstract.
* Allow @AutoAnnotation to work when the annotation and factory are in the default package.
* Don't generate the toBuilder constructor in AutoValue builders unless there actually is a toBuilder method.
* Treat property builders for nested @AutoValue types more like property builders for collections. If we have `ImmutableList<String> strings()` and `ImmutableList.Builder<String> stringsBuilder()`, then the list starts off empty. If you never add anything to it, the final result is an empty list. Likewise, if we have `NestedAutoValueType foo()` and `NestedAutoValueType.Builder fooBuilder()`, then `foo` starts off with its default value. If you never set anything in fooBuilder(), the final result is a default NestedAutoValueType. We only construct that default value lazily, so if you do in fact set something on the builder then the default instance of NestdAutoValueType is never constructed.
* In AutoValue builders, explicitly null-check the parameter of a setter when the corresponding property is not @Nullable. Previously we relied on the "null == missing" logic to trigger an exception in the build() method, but that turns out to be fragile, as well as being a less good experience.
* Update gwtserializer/pom.xml to run tests in prod mode.
* MoreTypes.asWildcard() should accept a type which isn't guaranteed to already be a WildcardType


value/src/it/gwtserializer/pom.xml
```diff
@@ -43,19 +43,17 @@
...
       <groupId>com.google.gwt</groupId>
       <artifactId>gwt-dev</artifactId>
-      <version>2.8.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>19.0</version>
+      <version>20.0</version>
```
ronshapiro: This doesn't have to be resolved here, but shoudn't this be `21`?

eamonnmcmanus: No, because that would introduce a Java 8 dependency. I'm planning to switch to requiring Java 8 after AutoValue 1.4 is released.

PR #444, opened 2017-03-08
Add support for Optional-wrapping property builders

octylFractal:
I think it is worthwhile to allow `Optional` for properties that use property builders, so I wrote code to support such an option.

The value is an empty `Optional` by default, and uses `Optional.of(builder.build())` to generate an `Optional` from the builder.

Ex:
```java
@AutoValue
abstract class Foo {
  @AutoValue.Builder
  interface Builder {
    Nested.Builder nestedBuilder();
  }
  
  abstract Optional<Nested> nested();
}

@AutoValue
abstract class Nested {
  
  static Builder builder() {
    return new AutoValue_Nested.Builder();
  }
  
  @AutoValue.Builder
  interface Builder {
  }
}
```


kenzierocks:
I've added the tests here as well, @eamonnmcmanus.

PR #450, opened 2017-03-15
Moe sync

ronshapiro:


eamonnmcmanus:
The Travis failure seems to be due to difference between the Java 7 and Java 8 compilers. On Java 7, the UNDEFINED error is output twice for some reason. Although we could fix the test to allow for that, I think we should just go ahead and release 1.4 without this change, then drop the Java 7 build. We'll be dropping support for building on Java 7 after 1.4.

ronshapiro:
Sounds good. Or you could just remove the warning count assertion, or `@Ignore` the test.

PR #457, opened 2017-03-20
Use badges for subproject release versions in README

hzsweers:


ronshapiro:
Can we replace these with badges instead? That way they auto-update and we don't have to continue to this periodically

hzsweers:
Good idea - done in 1ba4999 and 8c51565

PR #463, opened 2017-03-23
Fix link from read me to detailed documentation.

decbis:


decbis:
I signed it!

eamonnmcmanus:
Thanks! We're already aware of this problem and will be fixing it (everywhere it occurs) very soon.

PR #464, opened 2017-03-23
Change markdown links so they no longer have a line break between ] and (.

eamonnmcmanus:
Github recently stopped recognizing links like this, and pointed out <https://github.com/github/markup/issues/1026> that they are not covered by the Commonmark spec.

Some of these links might better be expressed using the link-reference style, [foo][bar] or just [foo], but for now I'm just making a whitespace change.

-------------
MOE_MIGRATED_REVID=150927275


eamonnmcmanus:
(Travis build currently broken on OpenJDK 7 for unrelated reasons.)

PR #468, opened 2017-03-28
We now require Java 8, so remove the @Ignore on annotationReferencesUndefinedTest

eamonnmcmanus:
-------------
MOE_MIGRATED_REVID=151405078


ronshapiro:
LGTM (assuming you figure out that failure)

eamonnmcmanus:
The failure is very strange. It's in a test that this PR doesn't touch, so I think it would probably have happened without the PR? I'm trying to see if I can reproduce it.

eamonnmcmanus:
I don't know what's up here and I can't reproduce it, even by making a very similar PR. So I'm going to merge this anyway and see what happens.

eamonnmcmanus:
Travis is passing now. We have some sort of transient problem, which *might* be fixed by adding `-implicit:none` to the various compilations in `CompilationTest.java`. If it starts occurring more reproducibly I will investigate further.

PR #470, opened 2017-03-30
Add detail to Effective Java reference.

clintonc:


eamonnmcmanus:
Thanks! We've copied this change into our internal version and synced it out in another PR.

PR #475, opened 2017-04-11
Remove iframe tag from md file

mlegy:
iframe tag is not supported in markdown, so it's shown like this

screen shot 2017-04-11 at 1 19 27 pm


eamonnmcmanus:
Thanks! The iframe does actually work on Google's internal systems, so I'll apply your fix there in a way that gets published correctly.

PR #476, opened 2017-04-11
typo correction in declared types equality check

doodeec:
There is a typo (originating as far as history of this repo remembers) in declared types equality check in `EqualVisitor`


eamonnmcmanus:
Oops. :-)
We'll make this change in Google's internal repo and it will appear (duly credited) when we sync out soon.

PR #477, opened 2017-04-18
Fix coding example

mbe24:
Add 'implements' to declaration of MyProcessor, since Processor cannot be extended by a class.
An alternative declaration would be `final class MyProcessor extends AbstractProcessor`.


eamonnmcmanus:
This was fixed in https://github.com/google/auto/pull/495.

PR #478, opened 2017-04-20
Update the Gradle instructions

vogella:
The previous instruction compileOnly and apt did not work for me in an new Android project created with latest Android Studio Beta.

### DIFF
--- a/value/userguide/index.md
+++ b/value/userguide/index.md
@@ -87,8 +87,8 @@ these instructions][tbroyer-apt] and then use it in the `build.gradle` script:
 
 ```groovy
 dependencies {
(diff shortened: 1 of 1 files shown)


jbduncan:
Do these new instructions work with bog standard Java builds?

jbduncan:
And for Android, is @tbroyer's apt plugin mentioned above in the source even needed?

devisnik:
@vogella The original instructions refer to integration into a regular Gradle java project. You might want to add android-specific ones instead of overriding them.

vogella:
Thanks for the feedback, I think it is better to close this request for now.

PR #480, opened 2017-04-21
Moe sync

ronshapiro:


value/src/main/java/com/google/auto/value/processor/BuilderSpec.java
```diff
@@ -131,10 +132,10 @@
...
         Types typeUtils, Set<ExecutableElement> abstractMethods) {
 
-      ImmutableList<String> builderTypeParamNames =
-          FluentIterable.from(builderTypeElement.getTypeParameters())
-              .transform(SimpleNameFunction.INSTANCE)
-              .toList();
+      List<String> builderTypeParamNames =
+          builderTypeElement.getTypeParameters().stream()
+              .map(e -> e.getSimpleName().toString())
+              .collect(toList());
```
jbduncan: There is an `ImmutableList.Builder` usage below (lines 145-150) that I believe could be replaced with streams as well (if one wants to go all out. :wink:)

eamonnmcmanus: Yes, I think I will do another pass over the code replacing for-loops with streams where appropriate.

PR #495, opened 2017-05-25
Improved example, Processor is an interface

arend-von-reinersdorff:


eamonnmcmanus:
Thanks! That example has been like that for years and nobody has noticed until now.

Our internal processes mean that we won't be merging this PR, but the change will appear in the sources in due course, duly attributed.

Issue #421, opened 2016-12-30
Proposal: Expose read-only builder properties in Extensions API


ZacSweers:
Hey all,

I've run into a few cases now where an extension implementation would be improved by being builder-aware (so it can generate code calling into it). What would you think of exposing this in a read-only fashion in the extensions API?

eamonnmcmanus:
That occurred to me recently as well. Since the class generated by an extension is a descendant of the AutoValue implementation, it inherits the generated Builder and can call it. My main concern is that builders are a lot more variable than what they build, so the API by which the contents of a builder might be exposed could be trickier. A minimal API would be just to tell the extension what the `@AutoValue.Builder` class is (to save it having to search), but then leave it to fend for itself when it comes to figure out how to build stuff. Things like `Optional` properties and nested builders mean this isn't necessarily trivial.

Could you describe some of the cases where this would have been useful?

hzsweers:
Here's an example: https://github.com/rharter/auto-value-gson/pull/89

eamonnmcmanus:
Suppose you had access to a `Map<String, Set<ExecutableElement>>` that told you what setters or builders there were for each property. Would that help? So for example given this...

```java
@AutoValue
abstract class Foo {
  abstract ImmutableList<String> strings();
  abstract Optional<Thread> thread();

  @AutoValue.Builder
  abstract static class Builder {
    abstract Builder setStrings(List<String> strings);
    abstract ImmutableList.Builder<String> stringsBuilder();
    abstract Builder setThread(Optional<Thread> thread);
    abstract Builder setThread(Thread thread);
    abstract Foo build();
  }
}
```

...the API would given you a Map where `"strings"` maps to the `setStrings` and `stringsBuilder` methods, and `"thread"` maps to the two `setThread` methods.

You'd still be a bit on your own to figure out what each kind of method is, and I'm afraid that extensions might be a bit fragile if we add new variants of setters that didn't exist when the extensions were written.

hzsweers:
hmm, yeah that could work. I think ideally it would give a reference to the `Builder` class itself, but not the worst thing in the world if they just required consumers to expose an entry point for instantiation in some "known location".

aurae:
[...]

PR #501, opened 2017-06-19
Adds read only builder properties to Extensions API

bryanstern:
This is a first pass implementation of issue #421, adding read only builder properties to the Extensions API.

The `AutoValueExtension` interface has been updated to overload the `generateClass` method to include an `Optional<BuilderContext>` parameter which is populated with information about the Builder, if there is one. The default implementation for this new method will call the older deprecated method if it is not overridden.

The `AutoValueExtension.Context` interface has been updated with a `hasBuilder` method. An Extension may require a Builder and should be informed by the `applicable` callback. We cannot provide a valid `BuilderContext` to the extension before the call `generateClass`, because we cannot validate the builder until all extensions have specified which methods they consume.


value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -550,7 +552,12 @@ private void processType(TypeElement type) {
     GwtCompatibility gwtCompatibility = new GwtCompatibility(type);
     vars.gwtCompatibleAnnotation = gwtCompatibility.gwtCompatibleAnnotationString();
 
-    int subclassDepth = writeExtensions(type, context, applicableExtensions);
+    Optional<BuilderContext> builderContext = Optional.empty();
+    if (builder.isPresent()) {
+      builderContext = builder.get().builderContext(typeSimplifier, properties.inverse());
```
JakeWharton: `map` instead of `isPresent`/`get`

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -785,6 +794,36 @@ private String extensionName(AutoValueExtension extension) {
...
+      TypeElement type,
+      ImmutableSet<ExecutableElement> propertyMethods,
+      Optional<BuilderSpec.Builder> builder) {
+    DeclaredType declaredType = MoreTypes.asDeclared(type.asType());
+    String pkg = TypeSimplifier.packageNameOf(type);
+
+    Set<TypeMirror> types = new TypeMirrorSet();
+    types.addAll(returnTypesOf(propertyMethods));
+    if (builder.isPresent()) {
+      types.addAll(builder.get().referencedTypes());
```
JakeWharton: `ifPresent(Consumer)` instead of `isPresent`/`get`

value/src/main/java/com/google/auto/value/processor/BuilderSpec.java
```diff
@@ -181,26 +184,37 @@
       return types;
     }
 
+    Optional<BuilderContext> builderContext(
+        TypeSimplifier typeSimplifier,
+        ImmutableBiMap<ExecutableElement, String> getterToPropertyName) {
+      Optional<BuilderMethodClassifier> optionalClassifier = builderMethodClassifier(typeSimplifier, getterToPropertyName);
```
JakeWharton: Whole method can be:
```java
return builderMethodClassifier(..).map(builder -> {
  stuff
});
```
instead of `ifPresent`/`get`

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
```diff
@@ -398,6 +409,102 @@ public void testExtensionWithBuilderCompilation() throws Exception {
...
+            "  @AutoValue.Builder public static abstract class Builder {",
+            "    public abstract Builder foo(String foo);",
+            "    public abstract Builder foo(Optional<String> foo);",
+            "    public abstract Builder bar(String bar);",
+            "    public abstract ImmutableList.Builder quxBuilder();",
+            "    abstract Optional<String> foo();",
+            "    abstract Baz autoBuild();",
+            "    public Baz build() {",
+            "      if (foo().isPresent()) {",
+            "        foo(foo().get() + \" fighters\");",
```
JakeWharton: Not that it matters here, but could also be `ifPresent(Consumer)` instead of `isPresent`/`get`.

value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
```diff
@@ -15,8 +15,11 @@
  */
 package com.google.auto.value.extension;
 
+import com.google.auto.value.AutoValue;
+
```
JakeWharton: nit: remove empty line per Google java format

value/src/main/java/com/google/auto/value/processor/BuilderMethodClassifier.java
```diff
@@ -44,12 +45,15 @@
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 
+import static com.google.auto.common.MoreElements.getLocalAndInheritedMethods;
```
JakeWharton: nit: static goes first per Google java format

value/src/main/java/com/google/auto/value/processor/BuilderMethodClassifier.java
```diff
@@ -163,9 +164,20 @@ private BuilderMethodClassifier(
...
   /**
-   * Returns the methods that were identified as {@code build()} methods. These are methods that
-   * have no parameters and return the {@code @AutoValue} type, conventionally called
-   * {@code build()}.
+   * Returns the abstract methods that were identified as {@code build()} methods. These are methods
+   * that have no parameters and return the {@code @AutoValue} type, conventionally called {@code
+   * build() or {@code autoBuild}}.
+   */
+  Set<ExecutableElement> abstractBuildMethods() {
+    return ImmutableSet.copyOf(buildMethods).stream()
```
JakeWharton: The `copyOf` seems superfluous here

value/src/main/java/com/google/auto/value/processor/BuilderMethodClassifier.java
```diff
@@ -163,9 +164,20 @@ private BuilderMethodClassifier(
   }
 
   /**
-   * Returns the methods that were identified as {@code build()} methods. These are methods that
-   * have no parameters and return the {@code @AutoValue} type, conventionally called
-   * {@code build()}.
+   * Returns the abstract methods that were identified as {@code build()} methods. These are methods
+   * that have no parameters and return the {@code @AutoValue} type, conventionally called {@code
+   * build() or {@code autoBuild}}.
```
JakeWharton: nit: missing `}` after `build()`

value/src/main/java/com/google/auto/value/processor/BuilderMethodClassifier.java
```diff
@@ -163,9 +164,20 @@ private BuilderMethodClassifier(
...
+  Set<ExecutableElement> abstractBuildMethods() {
+    return ImmutableSet.copyOf(buildMethods).stream()
+        .filter(MoreElements.hasModifiers(Modifier.ABSTRACT))
+        .collect(Collectors.toSet());
+  }
+
+  /**
+   * Returns the all methods that were identified as {@code build()} methods. These are methods
+   * that have no parameters, are public, and return the {@code AutoValue} type, conventionally
+   * called {@code build() or {@code autoBuild}}.
```
JakeWharton: nit: missing `}` after `build()`

value/src/main/java/com/google/auto/value/processor/ExtensionBuilderContext.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Sets;
+import java.util.Map;
+import java.util.Set;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+
+public class ExtensionBuilderContext implements AutoValueExtension.BuilderContext {
```
JakeWharton: -`public` +`final`

value/src/test/java/com/google/auto/value/processor/ExtensionTest.java
```diff
@@ -398,6 +409,102 @@ public void testExtensionWithBuilderCompilation() throws Exception {
   }
 
   @Test
+  public void testExtensionWithBuilder() throws Exception {
+    JavaFileObject javaFileObject = JavaFileObjects.forSourceLines(
+            "foo.bar.Baz",
```
JakeWharton: nit: continuation indent is 4 spaces

value/src/main/java/com/google/auto/value/processor/ExtensionBuilderContext.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  public Set<ExecutableElement> buildMethods() {
+    return buildMethods;
+  }
+
+  @Override
+  public Map<String, Set<ExecutableElement>> setters() {
+    return Maps.transformValues(setters.asMap(), Sets::newHashSet);
```
JakeWharton: Move to constructor? We already do a defensive copy might as well front-load this work instead.

bryanstern:
I signed it!

eamonnmcmanus:
Sorry I haven't been able to get to this. I will try to soon, and also start an internal API review for it. I will say that I think I prefer having an `Optional<BuilderContext>` getter in `Context` rather than a new overload for `generateClass`. The whole point of `Context` was to avoid having to change the signature of `generateClass` as the API evolves. I realize it is a bit ugly for the getter only to work within `generateClass` and not in the earlier phases, but I don't think it's as ugly as the deprecated overload.

bryanstern:
Sounds good. I'll see if I can come up with something clever. Also, currently we return a TypeElement for the abstract/interface of the builder. I think we should include one for the generated Builder so that extensions have to infer less about the Builder's constructor.

eamonnmcmanus:
I don't think that can work. When extensions run, the generated builder hasn't been compiled, so it can't have a TypeElement.

bryanstern:
Implemented your suggestion and realized my concern about the Builder TypeElement was not valid. Hopefully this is good to go.

bryanstern:
@eamonnmcmanus anything I can I do to push this forward?

eamonnmcmanus:
Actually, yes. If you could describe a motivating example when you would have wanted this, it would help me in describing the API change for review. #421 mentions https://github.com/rharter/auto-value-gson/pull/89 but I'm finding it hard to pull a description out of that. Presumably you're working on this because you want to use it somewhere.

rharter:
For rharter/auto-value-gson, and other extensions that offer generated creators of AutoValue types (via things like generated TypeAdapters), this feature allows the extension to utilize the users already defined default settings defined in the builder. 

For instance, a user might have a static `builder()` that returns a builder preconfigured with some default values. Extensions don't currently have visibility into the parsed builders to make use of that, so they have to find a way to reimplement that type of functionality, ultimately, in the case of auto-value-gson, passing that burden on to the user. In addition, it added a lot of overhead in method and variable count (important for Android) to have to reimplement this. 

Having visibility into the processed builder allows extensions to use the existing creation logic without added complexity for the user.

eamonnmcmanus:
Sorry if I'm a bit slow, but I'm not really following that example. This PR won't give extensions any help in finding the `builder()` method, right? I can see that it could be useful to auto-value-gson to be able to see what the builder class is and what the setters in that class are, but I don't really see the connection to default values or to code size. I understand that the extension might have to reimplement a lot of AutoValue's logic to find the builder and its setters, but I don't see how that impacts the size of the generated code.

bryanstern:
I would look at rharter/auto-value-gson#136 as an example. The most compelling reason is that it allows extensions to use the validation and normalization logic a user has already defined in the builder. An additional benefit of extensions utilizing the builder is that it leverages the convention of specifying defaults in a `builder` method, instead of each extension defining its own way of specifying defaults.

rharter:
@eamonnmcmanus The current implementation provides, via the context, the AutoValue parsed metadata about properties to extensions so that they don't have to reimplement the logic to parse the element to determine the properties vs. abstract or other methods.  This is useful to avoid a community of unreliable extensions, as it would be very hard to ensure all extensions correctly implement the logic, and would break compatibility with all extensions as soon as Auto changes anything regarding identifying properties.

The same cannot currently be said for Builders.  If an extension wants to use the AutoValue class's Builder, it has to reimplement all of the logic that's currently built into AutoValue, meaning there will almost certainly be small differences, missed edge cases, or inconsistency between extensions.

When I first wrote the extension support, I hadn't really considered builders, because I figured it wasn't difficult for generated code to call the many arg'ed constructors or static factory methods.  The use case I hadn't considered was default values.  In AutoValue, this can generally be accomplished in one of two ways: using multiple static factory methods, or using builders, which most non-trivial examples seem to prefer.

Any extension that wants to make use of default values, which could be any extension that interfaces with a format like JSON or XML that can contain optional properties, will likely want to allow the use to supply default values.  The current options for this are:

- simply don't support default values
- implementing your own runtime solution (the current auto-value-gson [solution](https://github.com/rharter/auto-value-gson#usage), which adds to the runtime overhead)
- reimplement AutoValue's builder logic.

The first two options are what appears to be commonly used in the community right now, as the third is quite a bit of overhead. 

This PR proposes supplying the Builder metadata to extensions, just like the AutoValue class metadata is already supplied, so that extensions don't need to provide fragile reimplementations.  

Another option to be to extract the code that processes the Builder to identify and match property setters, identify build methods, etc. into a separate artifact that extensions can depend on to ensure they use the same parsing logic.

eamonnmcmanus:
OK, let me see if I'm following. Today the typical way to handle default values is this:

```java
@AutoValue
abstract class Foo {
  abstract int bar();
  abstract int baz();
  static Builder builder() {
    return new AutoValue_Foo.Builder().setBar(23).setBaz(27);
  }

  @AutoValue.Builder
  abstract class Builder {
    abstract Builder setBar(int bar);
    abstract Builder setBaz(int baz);
    abstract Foo build();
  }
}
```

If an extension is going to call the builder with default values, then it will want to know what the setters for the builder are. So for example:

```java
abstract class Foo {
  @DefaultValue("17") abstract int bar();
  @DefaultValue("23") abstract int baz();
  static Builder fancyBuilder() {
    return AutoSomething_Foo.newBuilder();
  }
  ...
}
```

Here `AutoSomething_Foo` is generated by an extension, and it contains a `newBuilder()` method that returns a builder where properties are initialized to the value in the `@DefaultValue` annotation. In order to do that it has to be able to generate something like this:

```java
class AutoSomething_Foo {
  static Foo.Builder newBuilder() {
    return AutoValue_Foo.Builder().setBar(17).setBaz(23);
  }
}
```

The setters are `setBar(int)` and `setBaz(int)`, but they could also have been `bar(int)` and `baz(int)`. Today the extension would have to duplicate some of the logic already in `AutoValueProcessor` in order to tell which it is.

eamonnmcmanus:
I've been writing up a description of the API for internal review, and I'm finding it hard to justify having separate `hasBuilder()` and `builder()` methods. It seems more natural to have just one method, `java.util.Optional<BuilderContext> builder()`. I know there is a phasing problem, which is going to lead to an imperfect API one way or another, but I think it is better to return `Optional.of(builderContext)` where some methods in the `builderContext` can't be called until the `generateClass` phase. In particular it should be possible to call the `builderClass()` method, at least, from any phase.

rharter:
@eamonnmcmanus I think that's the general idea.  Other supporting use cases for this are validation and normalization.

For what it's worth, I think the case you provided, in which an extension is providing the defaults is less important (though your example might prove a good extension if people find manually writing and maintaining the code too hard, as it puts the default value with the property declaration).  I'm more thinking that an extension might be creating objects and wants to make use of the users custom logic.  Here's an example:

Json Input

```json
[
  {
    "name": "Dog",
    "numberOfLegs": 4
  },
  {
    "numberOfLegs": 3
  },
  {
    "name": "Human"
  }
]
```

Value Type

```java
@AutoValue
public abstract class Animal {
  public abstract String name();
  public abstract int numberOfLegs();

  public static Builder builder() {
    return new AutoValue_Animal.Builder().numberOfLegs(2);
  }

  @AutoValue.Builder
    public abstract Builder setName(String value);
    public abstract Builder setNumberOfLegs(int value);

    abstract Optional<String> name();
    abstract int numberOfLegs();

    abstract Animal autoBuild(); // not public

    public Animal build() {
      if (!name().isPresent()) {
        setName(numberOfLegs() + "-legged creature");
      } else {
        setName(name().get().toLowerCase());
      }
      return autoBuild();
    }
  }
}
```

In this example, without knowledge of the Builder and it's structure, an extension that parses the supplied JSON would not be able to produce valid Animals, based on the user's normalization.  The extension could reimplement the processing logic that AutoValue already performs, but that would be fraught with errors, especially for a complex example such as this.

eamonnmcmanus:
All right, but in this example the extension would still have to figure out for itself that it needed to use the `builder()` method. AutoValue doesn't know about `builder()` so the extension framework can't identify it for you. An extension would also need to guess which of `build()` or `autoBuild()` it should use, since it will get told about both of them.

(I think it's very helpful to work through this example.)

bryanstern:
Correct and I think that is ok. In my AutoValueGson PR we decided to that if the `build` or `builder` methods were overloaded, the user would need to annotate which method to use (e.g. `@AutoValueGsonBuild` and `@AutoValueGsonBuilder` respectively).

I think this makes sense for each extension manage this their own way because multiple extensions might be applied to one AutoValue'd class but would be interested in different `build` and `builder` methods.

rharter:
Exactly as @bryanstern mentioned, the extensions would be responsible for identifying which `builder()` method to use, but that would be trivial if we know the class type of the Builder. The logic would go something like this.

```
if (context says builder exists) {
  if (number of static, no-arg methods returning builder class == 1) {
    Use the static, no-arg method returning builder class
  } else if (one of the static, no-arg methods returning builder class is marked) {
    // i.e. with an extenstion specific annotation
    User the marked method
  } else {
    Revert to AutoValue_Foo constructor or (more likely) throw an error
  }
} else {
  Use the AutoValue_Foo constructor
}
```

As for the example above with multiple `build` methods in the builder, that would follow similar logic.  It would be trivial to identify build methods since we know the return type, though.

bryanstern:
Besides the API concerns brought up earlier, are there are any other blockers? I was waiting until the motivation for this work was approved before I started on the API changes.

eamonnmcmanus:
Sorry not to have been able to devote more time to this. I haven't finished writing up an API review proposal. I really don't expect the idea itself to be problematic, though some of the final API details might differ. I can probably make any tweaks necessary as I incorporate your code into our internal repo (duly credited of course). You can feel free either to make the API changes I suggested or leave things as they are and I'll make them later.

hzsweers:
@eamonnmcmanus any update on your end? This is actually a really powerful and useful API improvement, as a lot of libraries using extensions get pressure to add custom (awkward) handling for defaults that builders already help you provide

bryanstern:
@eamonnmcmanus, what is the status of this PR?

eamonnmcmanus:
Sorry to sit on this for so long. My plan is to make an API very close to this one and run it through Google's internal API review process, but I haven't found the time to do that. I will try to get to it soon.

hzsweers:
Did the internal review process land anywhere on this?

bryanstern:
Also curious to know where the review process landed on this.

bryanstern:
@eamonnmcmanus Could you an update on the project's plan for this functionality? We have some technical decisions to make on our project that would be better informed if we know that this will ship or not.

ZacSweers:
@eamonnmcmanus any update from your end? The added requirement that any autovalue that implements a builder has a private constructor means that extensions cannot call the super constructor in these cases.

> If you use the builder option, AutoValue will not generate a visible constructor for the generated concrete value class. If it's important to offer your caller the choice of a factory method as well as the builder, then your factory method implementation will have to invoke the builder itself.

valeraz:
@eamonnmcmanus One more ping to see if we could get this merged or whether an internal equivalent is still in the works. Totally understand if there's no bandwidth to land this, but it'd be great to have closure so that we know how to proceed with maintenance of our fork. Thanks!

eamonnmcmanus:
Sorry this has taken so very long. I'm planning to take this up in the near future.

efraimg:
> Sorry this has taken so very long. I'm planning to take this up in the near future.

Waiting for it eagerly, I ended up writing my own serializers/de-serializers :(

eamonnmcmanus:
We've implemented something quite close to this in https://github.com/google/auto/commit/832480ea205c1f874d972729b74f1362303756e9. Thanks @bryanstern for the original proposal, which was the starting point for the change.

PR #507, opened 2017-07-02
Explicitly narrow long to int before adding to hash.

### LINKED ISSUE #506: ErrorProne: [NarrowingCompoundAssignment] Compound assignments from long to int hide lossy casts
Starting with Error Prone 2.0.20 some of the generated Auto Value code gets flagged:

```
warning: [NarrowingCompoundAssignment] Compound assignments from long to int hide lossy casts
    h ^= (this.id >>> 32) ^ this.id;
      ^
    (see http://errorprone.info/bugpattern/NarrowingCompoundAssignment)
```

`id` is defined as `abstract int id()`.

Is this really an issue or can AutoValue guarantee that this will work?

JakeWharton:
The compound assignment to the hash value already performed the integer narrowing but did so implicitly. Error-prone was recently updated to detect this case and thus marked this as problematic. Rasterizing the implicit cast as explicitly makes the intent clear and resolves the warnings.

This behavior is already present in AutoAnnotation but was missing from AutoValue.

Closes #506.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/autovalue.vm
+++ b/value/src/main/java/com/google/auto/value/processor/autovalue.vm
@@ -147,11 +147,11 @@ $a
     #if ($p.kind == "BYTE" || $p.kind == "SHORT" || $p.kind == "CHAR" || $p.kind == "INT")
       this.$p ##
     #elseif ($p.kind == "LONG")
-      (this.$p >>> 32) ^ this.$p ##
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
Thanks! This is indeed the obvious fix. As you probably know, our processes mean that we don't merge PRs, but I'll make an equivalent change internally (along with a small test).

PR #515, opened 2017-08-21
Fix generics issue with builder-based AutoValues.

perlow:
This fixes issues with having an AutoValue-with-builder class that is itself a parameterized subclass of a base class with a generic parameter. Previously, the AutoValue code was using the type variable as the parameter in the AutoValue instead of the resolved type.


eamonnmcmanus:
Sorry that I appear to have missed this when it was created. We've avoided using Types.asMemberOf because of very longstanding bugs in the Eclipse compiler. If there's reason to think those bugs have been fixed we could reconsider.

BruceBGordon:
+1 for this pr. We're also have issues with Builders and generics.

eamonnmcmanus:
Can someone check whether the patched AutoValue works with Eclipse?

Meanwhile, I'll mention again that there is a workaround for the problem: just restate the method in question with the correct type. Bug https://github.com/google/auto/issues/627 illustrates this workaround. I realize that this is annoying, though.

chrisbouchard:
@eamonnmcmanus I know it's been a while. I'm not part of the project, but I love Auto/Value and I'd be willing to try this out in Eclipse in the near future (if I can get everything working). Do you have any thoughts how you'd expect bugs with `Types.asMember` to manifest?

eamonnmcmanus:
@chrisbouchard that would be much appreciated!

In the past there have been problems when the target type was an interface. So looking at the [test code](https://github.com/google/auto/pull/515/files#diff-8247d4bb64f54e813b09c00139caed9eR2718) from this PR, you could try changing the static class to an interface, like this:

```java
package foo.bar;

public abstract class BaseClass<T> {

  public abstract T value();

  public interface Builder<BUILDER extends Builder<BUILDER, T>, T> {
    public abstract T value();
    public abstract BUILDER value(T value);
  }
}
```

and

```java
package foo.bar;
import com.google.auto.value.AutoValue;

@AutoValue
public abstract class ExtendedClass extends BaseClass<Integer> {

  @AutoValue.Builder
  public interface Builder extends BaseClass.Builder<Builder, Integer> {
    public abstract ExtendedClass build();
  }
}
```

If that works in Eclipse it should prove both that you are using the patched version and that it doesn't trigger the specific Eclipse problem I had in mind.

eamonnmcmanus:
This is obsolete. I thoroughly reworked the way that types are handled in builders and, as far as I know, all of the many outstanding problems were resolved at that point.

PR #520, opened 2017-08-30
Sync 2017 08 29

eamonnmcmanus:


ronshapiro:
@eamonnmcmanus sorry I missed this, but I presume by now it's outdated? Though

eamonnmcmanus:
Actually no, I don't see these commits, so I think this is still relevant.

eamonnmcmanus:
Never mind, it's older than I thought.

PR #522, opened 2017-09-05
Ignore IntelliJ IDEA project files when not stored in .idea directory

realityforge:


realityforge:
Closing this as unlikely to be applied

Issue #449, opened 2017-03-15
Require Java 8 for running AutoValue


eamonnmcmanus:
Up until now we have ensured that AutoValue would compile and run on Java 6 and would generate code that itself compiles on Java 6. In https://github.com/google/auto/issues/448 I am proposing that we drop the requirement that the generated code can compile on Java 6, and instead target Java 7. However, generating code that can run on Java 7 does not require that it can be *compiled* on Java 7. It can perfectly well be compiled on a later platform with `-source 7 -target 7 -bootclasspath <rt.jar-from-7>`. I am therefore proposing that, after AutoValue 1.4, we allow the source code of AutoValue itself to use Java 8 fully.

eamonnmcmanus:
This requirement is now in place.

bogdandrutu:
Hey @eamonnmcmanus I think I may understand the reason behind moving to java7 from java6, but jumping to java8 will make any product/library to not be able to run on Android at all. Android still has to support a lot of Java7 devices. What is the story for that?

eamonnmcmanus:
You can still compile code with `-source 7 -target 7` so that it can run on Android. The only change is that the *compiler* you use must be a Java 8 compiler. That should be the case for all modern development environments, including Android ones.

rkraneis:
You might want to mention that in the Readme. It is not immediately apparent, that the generated code will be Java 7 compliant and compilable with these source/target settings.

PR #524, opened 2017-09-07
Update stale references from Java 1.6 to 1.7

gaul:
References #449.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Save time.  Save code.  Save sanity.
 
     [![Maven Central](https://img.shields.io/maven-central/v/com.google.auto.service/auto-service.svg)](https://mvnrepository.com/artifact/com.google.auto.service/auto-service)
 
-  * [AutoValue] - Immutable [value-type] code generation for Java 1.6+.
+  * [AutoValue] - Immutable [value-type] code generation for Java 1.7+.
 
     [![Maven Central](https://img.shields.io/maven-central/v/com.google.auto.value/auto-value.svg)](https://mvnrepository.com/artifact/com.google.auto.value/auto-value)
 
--- a/value/README.md
+++ b/value/README.md
@@ -1,6 +1,6 @@
 # AutoValue
 
-*Generated immutable value classes for Java 1.6+* <br />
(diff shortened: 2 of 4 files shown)


andrewgaul:
It would also be nice to document this new requirement in the release notes:

https://github.com/google/auto/releases/tag/auto-value-1.5

eamonnmcmanus:
The suggestion about the release notes is good and I have followed it.

There's a certain ambiguity about what we mean when we say that this is "for" a certain version of Java. It is true that the compiler must now be at least a Java 8 compiler. However it can still be used to generate Java 7 code, and I expect people to continue doing that for quite some time on Android. I think it might be best to delete these places where we say "for" some version, and instead add a note about versions to the user guide. I think it is reasonable to assume that nearly all users are on Java 8 tooling at this point, and those who are not should already be used to not everything being available to them.

Did you run into the Java 8 requirement yourself? What was the context?

JakeWharton:
The class files in 1.5 use version 52 (Java 8) which means a JDK 7 compiler
won't be able to use the annotation I believe.

rharter:
Correct, you'll get this if you try: https://github.com/rharter/auto-value-gson/issues/142

eamonnmcmanus:
@JakeWharton I didn't think I said anything different.

JakeWharton:
The change in retention means much more of your toolchain is impacted. Retaining the auto value class files on a classpath to appease tools looking at such things and the lack of split between annotation and compiler means the Java 7 thing is only _mostly_ accurate.

JakeWharton:
Starting with version 3.0 of the Android plugin we actually have a nice get-out-of-jail-free card where the whole toolchain now supports Java 8 class files to make this a non-problem. But it's currently opt-in ☹️

eamonnmcmanus:
Interesting. The case for separate artifacts is certainly stronger now, since that would make it easier for AutoValue.class to be a Java 7 class file.

andrewgaul:
Thanks for updating the release notes -- this reflects my experience testing locally.  I updated the pull request to Java 1.7 but please close it if you plan a larger explanation based on this discussion.  For what it is worth, I am testing with Apache jclouds which still supports Java 1.6.

PR #532, opened 2017-09-28
Moe sync

ronshapiro:


eamonnmcmanus:
Looks as if GwtCompilationTest.java isn't exported. I'll look into why. Meanwhile there's nothing wrong with this PR as such.

PR #548, opened 2017-11-13
Proposal PR: Change AutoService#value() to Class[]

### LINKED ISSUE #488: @AutoService value: accept array of class instead of single class
I think it would make sense to be able to implement more than one service by the same class.
(would require java 8+ though)

hzsweers:
This resolves #488

This changes the value() type in AutoService to be Class[], allowing for consumers to declare support for multiple service interfaces. This is a backward compatible change in source while allowing for greater flexibility. See `MultiServiceProvider.java` for an example.

### DIFF
--- a/service/src/main/java/com/google/auto/service/AutoService.java
+++ b/service/src/main/java/com/google/auto/service/AutoService.java
@@ -40,6 +40,6 @@ import java.lang.annotation.Target;
 @Retention(SOURCE)
 @Target(TYPE)
 public @interface AutoService {
-  /** Returns the interface implemented by this service provider. */
-  Class<?> value();
+  /** Returns the interfaces implemented by this service provider. */
+  Class<?>[] value();
 }
--- a/service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
+++ b/service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
@@ -17,6 +17,7 @@ package com.google.auto.service.processor;
 
 import static com.google.auto.common.MoreElements.getAnnotationMirror;
 
+import com.google.auto.common.MoreTypes;
 import com.google.auto.service.AutoService;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
@@ -141,7 +142,7 @@ public class AutoServiceProcessor extends AbstractProcessor {
         continue;
       }
       for (DeclaredType providerInterface : providerInterfaces) {
-        TypeElement providerType = (TypeElement) providerInterface.asElement();
+        TypeElement providerType = MoreTypes.asTypeElement(processingEnv.getTypeUtils(), providerInterface);
 
         log("provider interface: " + providerType.getQualifiedName());
         log("provider implementer: " + providerImplementer.getQualifiedName());
--- a/service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
+++ b/service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
@@ -15,24 +15,23 @@
  */
 package com.google.auto.service.processor;
 
-import static com.google.auto.common.MoreElements.getAnnotationMirror;
-
+import com.google.auto.common.AnnotationMirrors;
 import com.google.auto.common.MoreTypes;
 import com.google.auto.service.AutoService;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
+
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.lang.annotation.Annotation;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
+
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.Filer;
 import javax.annotation.processing.RoundEnvironment;
@@ -41,16 +40,16 @@ import javax.lang.model.SourceVersion;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
-import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
-import javax.lang.model.type.TypeMirror;
-import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 import javax.tools.Diagnostic.Kind;
 import javax.tools.FileObject;
 import javax.tools.StandardLocation;
+
+import static com.google.auto.common.AnnotationMirrors.getAnnotationValue;
+import static com.google.auto.common.MoreElements.getAnnotationMirror;
 
 /**
  * Processes {@link AutoService} annotations and generates the service provider
@@ -133,15 +132,13 @@ public class AutoServiceProcessor extends AbstractProcessor {
       // TODO(gak): check for error trees?
       TypeElement providerImplementer = (TypeElement) e;
       AnnotationMirror annotationMirror = getAnnotationMirror(e, AutoService.class).get();
-      Set<DeclaredType> providerInterfaces = getFieldOfClasses(
-          annotationMirror,
-          "value");
+      Set<DeclaredType> providerInterfaces = getValueFieldOfClasses(annotationMirror);
       if (providerInterfaces.isEmpty()) {
         error("No interfaces provided for element!", e, annotationMirror);
         continue;
       }
       for (DeclaredType providerInterface : providerInterfaces) {
-        TypeElement providerType = MoreTypes.asTypeElement(processingEnv.getTypeUtils(), providerInterface);
+        TypeElement providerType = MoreTypes.asTypeElement(providerInterface);
 
         log("provider interface: " + providerType.getQualifiedName());
         log("provider implementer: " + providerImplementer.getQualifiedName());
@@ -252,7 +249,7 @@ public class AutoServiceProcessor extends AbstractProcessor {
   }
 
   /**
-   * Returns the contents of a {@code Class[]}-typed field in a given {@code annotationMirror}.
+   * Returns the contents of a {@code Class[]}-typed "value" field in a given {@code annotationMirror}.
    *
    * <p>This method is needed because directly reading the value of such a field from an
    * AnnotationMirror throws: <pre>
@@ -260,28 +257,15 @@ public class AutoServiceProcessor extends AbstractProcessor {
    * </pre>
    *
    * @param annotationMirror The AnnotationMirror to read fields from, e.g. {@link AutoService}.
-   * @param fieldName The name of the field to read, e.g. "value".
    * @return a set of fully-qualified names of classes appearing in 'fieldName' on 'annotation' on
    *     'element'.
    */
-  private ImmutableSet<DeclaredType> getFieldOfClasses(
-      AnnotationMirror annotationMirror,
-      String fieldName) {
-    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry :
-        annotationMirror.getElementValues().entrySet()) {
-      if (fieldName.contentEquals(entry.getKey().getSimpleName())) {
-        ImmutableSet.Builder<DeclaredType> result = ImmutableSet.builder();
-
-        @SuppressWarnings("unchecked")
-        List<AnnotationValue> annotationsToCopy =
-            (List<AnnotationValue>) entry.getValue().getValue();
-        for (AnnotationValue annotationValue : annotationsToCopy) {
-          result.add((DeclaredType) annotationValue.getValue());
-        }
-        return result.build();
-      }
-    }
-    return ImmutableSet.of();
+  private ImmutableSet<DeclaredType> getValueFieldOfClasses(AnnotationMirror annotationMirror) {
+    //noinspection unchecked
+    return ((List<AnnotationValue>) getAnnotationValue(annotationMirror, "value").getValue())
+            .stream()
+            .map(value -> MoreTypes.asDeclared((DeclaredType) value.getValue()))
+            .collect(ImmutableSet.toImmutableSet());
   }
 
   private void log(String msg) {
--- a/service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
+++ b/service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
@@ -26,7 +26,9 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.annotation.Annotation;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
@@ -42,6 +44,8 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.PackageElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 import javax.tools.Diagnostic.Kind;
 import javax.tools.FileObject;
@@ -127,26 +131,32 @@ public class AutoServiceProcessor extends AbstractProcessor {
     for (Element e : elements) {
       // TODO(gak): check for error trees?
       TypeElement providerImplementer = (TypeElement) e;
-      AnnotationMirror providerAnnotation = getAnnotationMirror(e, AutoService.class).get();
-      DeclaredType providerInterface = getProviderInterface(providerAnnotation);
-      TypeElement providerType = (TypeElement) providerInterface.asElement();
+      AnnotationMirror annotationMirror = getAnnotationMirror(e, AutoService.class)
+          .get(); // We already know it's present
+      Set<DeclaredType> providerInterfaces = getFieldOfClasses(
+          annotationMirror,
+          "value",
+          processingEnv.getElementUtils());
+      for (DeclaredType providerInterface : providerInterfaces) {
+        TypeElement providerType = (TypeElement) providerInterface.asElement();
 
-      log("provider interface: " + providerType.getQualifiedName());
-      log("provider implementer: " + providerImplementer.getQualifiedName());
+        log("provider interface: " + providerType.getQualifiedName());
+        log("provider implementer: " + providerImplementer.getQualifiedName());
 
-      if (!checkImplementer(providerImplementer, providerType)) {
-        String message = "ServiceProviders must implement their service provider interface. "
-            + providerImplementer.getQualifiedName() + " does not implement "
-            + providerType.getQualifiedName();
-        error(message, e, providerAnnotation);
+        if (!checkImplementer(providerImplementer, providerType)) {
+          String message = "ServiceProviders must implement their service provider interface. "
+              + providerImplementer.getQualifiedName() + " does not implement "
+              + providerType.getQualifiedName();
+          error(message, e, annotationMirror);
+        }
+
+        String providerTypeName = getBinaryName(providerType);
+        String providerImplementerName = getBinaryName(providerImplementer);
+        log("provider interface binary name: " + providerTypeName);
+        log("provider implementer binary name: " + providerImplementerName);
+
+        providers.put(providerTypeName, providerImplementerName);
       }
-
-      String providerTypeName = getBinaryName(providerType);
-      String providerImplementerName = getBinaryName(providerImplementer);
-      log("provider interface binary name: " + providerTypeName);
-      log("provider implementer binary name: " + providerImplementerName);
-
-      providers.put(providerTypeName, providerImplementerName);
     }
   }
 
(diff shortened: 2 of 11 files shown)


service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -127,26 +131,32 @@ private void processAnnotations(Set<? extends TypeElement> annotations,
...
-            + providerType.getQualifiedName();
-        error(message, e, providerAnnotation);
-      }
+      AnnotationMirror annotationMirror = getAnnotationMirror(e, AutoService.class)
+          .get(); // We already know it's present
+      Set<DeclaredType> providerInterfaces = getFieldOfClasses(
+          annotationMirror,
+          "value",
+          processingEnv.getElementUtils());
+      for (DeclaredType providerInterface : providerInterfaces) {
```
ronshapiro: We're gonna want to validate that `!providerInterfaces.isEmpty()`

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -127,26 +131,32 @@ private void processAnnotations(Set<? extends TypeElement> annotations,
...
+        TypeElement providerType = (TypeElement) providerInterface.asElement();
```
ronshapiro: I know this is a move, but might as well do some tiding up while we're here:
`MoreTypes.asTypeElement(providerInterface)`

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -127,26 +131,32 @@ private void processAnnotations(Set<? extends TypeElement> annotations,
...
+
+        log("provider interface: " + providerType.getQualifiedName());
+        log("provider implementer: " + providerImplementer.getQualifiedName());
+
+        if (!checkImplementer(providerImplementer, providerType)) {
```
ronshapiro: What about:

```java
if (checkImplementer(...)) {
  providers.put(...);
} else {
  error(...);
}
```

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -127,26 +131,32 @@ private void processAnnotations(Set<? extends TypeElement> annotations,
...
+          error(message, e, annotationMirror);
+        }
 
-      String providerTypeName = getBinaryName(providerType);
-      String providerImplementerName = getBinaryName(providerImplementer);
-      log("provider interface binary name: " + providerTypeName);
-      log("provider implementer binary name: " + providerImplementerName);
+        String providerTypeName = getBinaryName(providerType);
+        String providerImplementerName = getBinaryName(providerImplementer);
+        log("provider interface binary name: " + providerTypeName);
```
ronshapiro: Can probably get rid of these logs, I doubt they're doing a whole lot.

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +253,38 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
+   * @param annotationMirror The AnnotationMirror to read fields from, e.g. {@link AutoService}.
+   * @param fieldName The name of the field to read, e.g. "value".
+   * @return a set of fully-qualified names of classes appearing in 'fieldName' on 'annotation' on
+   *     'element'.
+   */
+  private ImmutableSet<DeclaredType> getFieldOfClasses(
+      AnnotationMirror annotationMirror,
+      String fieldName,
+      Elements elementUtils) {
+    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry :
```
ronshapiro: I think this can be:

```java
((List<AnnovationValue>) AnnotationMirrors.getAnnotationValue(annotationMirror, "value"))
    .stream()
    .map(value -> MoreTypes.asDeclared(value.getValue())
    .collect(toImmutableSet());
```

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +253,38 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
+   *
+   * @param annotationMirror The AnnotationMirror to read fields from, e.g. {@link AutoService}.
+   * @param fieldName The name of the field to read, e.g. "value".
+   * @return a set of fully-qualified names of classes appearing in 'fieldName' on 'annotation' on
+   *     'element'.
+   */
+  private ImmutableSet<DeclaredType> getFieldOfClasses(
+      AnnotationMirror annotationMirror,
+      String fieldName,
+      Elements elementUtils) {
```
ronshapiro: This is unused

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -127,26 +131,32 @@ private void processAnnotations(Set<? extends TypeElement> annotations,
...
-      log("provider interface: " + providerType.getQualifiedName());
-      log("provider implementer: " + providerImplementer.getQualifiedName());
-
-      if (!checkImplementer(providerImplementer, providerType)) {
-        String message = "ServiceProviders must implement their service provider interface. "
-            + providerImplementer.getQualifiedName() + " does not implement "
-            + providerType.getQualifiedName();
-        error(message, e, providerAnnotation);
-      }
+      AnnotationMirror annotationMirror = getAnnotationMirror(e, AutoService.class)
```
ronshapiro: I think you can revert this line change, the comment doesn't do a whole lot.

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
```diff
@@ -38,6 +38,7 @@ public void autoService() {
             JavaFileObjects.forResource("test/SomeServiceProvider1.java"),
             JavaFileObjects.forResource("test/SomeServiceProvider2.java"),
             JavaFileObjects.forResource("test/Enclosing.java"),
+            JavaFileObjects.forResource("test/MultiServiceProvider.java"),
```
ronshapiro: I think this should be a separate test with just itself and `SomeService`+`AnotherService` to make sure that both get generated. Either that, or we should get the contents and assert that they have what we expect.

hzsweers:
Oh woah - I did not get notifications that this had been reviewed. Will work on it this week!

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: Good point. Done in d05dca4

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: Done (with some extras) in dfbbdd4

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: 6b870da

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: b68e691

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: I thought about that, but concerned the behavior change (even though it seems like it's obviously wrong) might be a bit significant if people for some reason relied on it. Done in 7a70c82

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: 7578b71 (and updated for the new API in dfbbdd4 after updating guava)

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: 5807e71 and a test in d05dca4. Ignore the commit message, I can't boolean

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: ac5a649

hzsweers:
Addressed existing comments. Regarding the test failure - my maven-fu isn't that strong, but I manually verified it does generate the expected classes. What's stranger is that the existing test now fails if I remove `MultiServiceProvider` 🤔 

Any thoughts?

service/src/main/java/com/google/auto/service/AutoService.java
```diff
@@ -40,6 +40,6 @@
 @Retention(SOURCE)
 @Target(TYPE)
 public @interface AutoService {
-  /** Returns the interface implemented by this service provider. */
-  Class<?> value();
+  /** Returns the interfaces implemented by this service provider. */
```
ronshapiro: Maybe `interface(s)`

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +250,24 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
-    Map<? extends ExecutableElement, ? extends AnnotationValue> valueIndex =
-        providerAnnotation.getElementValues();
-    log("annotation values: " + valueIndex);
-
-    AnnotationValue value = valueIndex.values().iterator().next();
-    return (DeclaredType) value.getValue();
+  /**
+   * Returns the contents of a {@code Class[]}-typed "value" field in a given {@code annotationMirror}.
+   *
+   * <p>This method is needed because directly reading the value of such a field from an
```
ronshapiro: I don't think this block is necessary - this is a common annotation-processing topic.

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +250,24 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
+   * AnnotationMirror throws: <pre>
+   * javax.lang.model.type.MirroredTypeException: Attempt to access Class object for TypeMirror Foo.
+   * </pre>
+   *
+   * @param annotationMirror The AnnotationMirror to read fields from, e.g. {@link AutoService}.
```
ronshapiro: I don't think these tags are necessary either. The first one is obvious to me and the second, and the second one is out of date at least. The first sentence of the javadoc makes both of them clear.

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +250,24 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
+   * @return a set of fully-qualified names of classes appearing in 'fieldName' on 'annotation' on
+   *     'element'.
+   */
+  private ImmutableSet<DeclaredType> getValueFieldOfClasses(AnnotationMirror annotationMirror) {
+    //noinspection unchecked
+    return ((List<AnnotationValue>) getAnnotationValue(annotationMirror, "value").getValue())
+            .stream()
+            .map(value -> MoreTypes.asDeclared((DeclaredType) value.getValue()))
```
ronshapiro: Either cast to `(TypeMirror)` and pass to `MoreTypes.asDeclared()` or just cast to `DeclaredType`.

Can you add this todo for me:

 `TODO(ronshapiro): class literals may not always be declared types, i.e. int.class, int[].class`

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +250,24 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
+            .collect(ImmutableSet.toImmutableSet());
```
ronshapiro: static import `toImmutableSet`

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
```diff
@@ -46,4 +48,30 @@ public void autoService() {
             JavaFileObjects.forResource("META-INF/services/test.SomeService"),
             JavaFileObjects.forResource("META-INF/services/test.AnotherService"));
   }
-}
\ No newline at end of file
+
+  @Test
+  public void multiService() {
+    assert_().about(javaSources())
```
ronshapiro: `JavaSourcesSubjectFactory` has a `static` `assertThat` method that accepts a varargs.

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
```diff
@@ -46,4 +48,30 @@ public void autoService() {
...
+        .that(Arrays.asList(
+            JavaFileObjects.forResource("test/SomeService.java"),
+            JavaFileObjects.forResource("test/MultiServiceProvider.java"),
```
ronshapiro: Can you switch this line with the next?

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
```diff
@@ -46,4 +48,30 @@ public void autoService() {
...
+        .and().generatesFiles(
+            JavaFileObjects.forResource("META-INF/services/test.SomeService"),
+            JavaFileObjects.forResource("META-INF/services/test.AnotherService"));
+  }
+
+  @Test
+  public void badMultiService() {
+    assert_().about(javaSources())
+        .that(Arrays.asList(
+            JavaFileObjects.forResource("test/SomeService.java"),
```
ronshapiro: You could remove SomeService and AnotherService, and have BadMultiServiceProvider not implement anything.

ronshapiro:
FYI @eamonnmcmanus this is getting close - do you want to add in a review here and then we can merge internally?

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
```diff
@@ -46,4 +48,30 @@ public void autoService() {
...
+  public void multiService() {
+    assert_().about(javaSources())
+        .that(Arrays.asList(
+            JavaFileObjects.forResource("test/SomeService.java"),
+            JavaFileObjects.forResource("test/MultiServiceProvider.java"),
+            JavaFileObjects.forResource("test/AnotherService.java")))
+        .processedWith(new AutoServiceProcessor())
+        .compilesWithoutError()
+        .and().generatesFiles(
+            JavaFileObjects.forResource("META-INF/services/test.SomeService"),
```
ronshapiro: I think you'd want these to be a different file from the above test's file.

Alternatively, we can probably use `JavaFileObjects.forSourceLines()` for both so we don't need to rely on the actual paths of a checked-in-file

service/src/main/java/com/google/auto/service/AutoService.java
eamonnmcmanus: I think `interfaces` is better. I don't know how to pronounce `interface(s)`.

ronshapiro: Are you saying that the suffix (s) is not common English?

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: I'm not seeing this, can you link me where that's defined?

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: I don't quite follow the first comment? Todo is here: 9f81d2a

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: 1c2a510

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: dc43d90

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: bbbd770

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: 712d0ae

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: d2d4c4c

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: problem seems to be deeper. I misunderstood the error message, and it seems the issue is that it's not finding the generated file, not the expected ones to compare them to. Wondering if it has to do with the compile-testing update

service/src/main/java/com/google/auto/service/AutoService.java
hzsweers: I defer to whoever has the more assertive thesaurus

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
ronshapiro: Whoops, I meant `JavaSourcesSubject` (not the factory)

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
ronshapiro: `MoreTypes.asDeclared()` is essentially a cast to a `DeclaredType`, so passing in something that's already declared should do nothing. so you can replace the line with this:
`.map(value -> (DeclaredType) value.getValue())`

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
ronshapiro: Nit: maybe rename `BadMultiService` to `NoServices`?

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
ronshapiro: Try using the `Compilation` and `CompilationSubject` apis? Those might be easier to use for this

service/src/main/java/com/google/auto/service/AutoService.java
eamonnmcmanus: I am saying that it is not good English. In the context, there's absolutely no reason any reader would think that the plural implies there must be more than one interface, and there is therefore absolutely no reason to pull readers up short with the unpronounceable `interface(s)`.

You might guess that this is a hobby horse of mine. :-)

hzsweers: The reason(s) above make sense to me

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: There's no API in compile-testing's clauses to get the underlying `Compilation` from what I can see. Do you have an example of what you mean?

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
hzsweers: 236f337

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: 94705d7

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: e4c030f

hzsweers:
So I've stepped through the debugger and confirmed the underlying `compilation` does in fact have the generated files, but it's failing this check in compile-testing:

```
expectedByteSource.contentEquals(JavaFileObjects.asByteSource(generated))
```

From https://github.com/google/compile-testing/blob/master/src/main/java/com/google/testing/compile/JavaSourcesSubject.java#L469

My guava-fu is not that strong, but something seems off here. This fails with both reading from resources and from having the code inlined directly in the test via forSourceLines. Am I missing something here?

ronshapiro:
I can't find the exact comment, but when you were asking where to find `Compilation`, the pattern goes like this:

```java
import static com.google.testing.compile.Compiler.javac;

Compilation compilation = javac().compile(files);
assertThat(compilation).succeeded();
assertThat(compilation)
    .generatedFile("META-INF/services/foo.bar.baz")
    .contentsAsString(StandardCharsets.UTF_8)
    .isEqualTo(Joiner.on('\n').join(line 1, line 2, line 3, etc));
```

hzsweers:
Finally fixed the tests... awkward but I was really thrown off by the error messages. It was finding the files, but the error messages are vague when emitted. The problem was that I needed to have separate services files for multiservice and needed to remove its references from the first test

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
```diff
@@ -243,23 +251,16 @@ private String getBinaryNameImpl(TypeElement element, String className) {
...
-        providerAnnotation.getElementValues();
-    log("annotation values: " + valueIndex);
-
-    AnnotationValue value = valueIndex.values().iterator().next();
-    return (DeclaredType) value.getValue();
+  /**
+   * Returns the contents of a {@code Class[]}-typed "value" field in a given {@code annotationMirror}.
+   */
+  private ImmutableSet<DeclaredType> getValueFieldOfClasses(AnnotationMirror annotationMirror) {
+    //noinspection unchecked
```
ronshapiro: nit: remove

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
```diff
@@ -15,35 +15,54 @@
...
 /**
  * Tests the {@link AutoServiceProcessor}.
  */
 @RunWith(JUnit4.class)
 public class AutoServiceProcessorTest {
   @Test
   public void autoService() {
-    assert_().about(javaSources())
-        .that(Arrays.asList(
+      JavaSourcesSubject.assertThat(
```
ronshapiro: nit: Can you static import `assertThat`

eamonnmcmanus:
LGTM

service/src/test/java/com/google/auto/service/processor/AutoServiceProcessorTest.java
hzsweers: bba6fa7

service/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java
ronshapiro: Ah, I didn't realize what you were doing. I'll just change it to a regular suppress warnings

hzsweers: :+1:

PR #556, opened 2017-11-29
Set up auto-parent with dependency and plugin management for the othe…

netdpb:
…r projects.

Update dependency and plugin versions.

Once this is released as auto-parent v5, the other auto* poms can start referring to it.

-------------
MOE_MIGRATED_REVID=177360283


eamonnmcmanus:
LGTM (sitting beside @netdpb).

PR #561, opened 2017-12-06
Moe sync

ronshapiro:


eamonnmcmanus:
Do we want to merge this? Won't it be undone shortly anyway?

ronshapiro:
I'm happy to wait if we'd like

PR #577, opened 2018-01-05
Change NullPointerException to IllegalArgumentException

weidox:
NullPointerException is for null reference access, not for parameter validation, and in general should not be throwed by user code. Use some other exception, like IllegalArgumentException or IllegalStateException depending on circumstances.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/autovalue.vm
+++ b/value/src/main/java/com/google/auto/value/processor/autovalue.vm
@@ -55,10 +55,10 @@ $a
     #if ($identifiers)
 
     if ($p == null) {
-      throw new NullPointerException("Null $p.name");
+      throw new IllegalArgumentException("Null $p.name");
     }
     #else
-      ## Just throw NullPointerException with no message if it's null.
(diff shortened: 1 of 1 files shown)


weidox:
I signed CLA.

cpovirk:
There's some debate about this here: https://stackoverflow.com/q/3881/28465

I'm pretty sure that we're happy following the precedent of the JDK and the advice in Effective Java (Item 60 in the 2nd edition), which suggest throwing `NullPointerException`, as permitted in [its docs](https://docs.oracle.com/javase/8/docs/api/java/lang/NullPointerException.html).

weidox:
Interesting. One of checkstyle/pmd/findbugs validators (don't remember which one) is checking against usage of NPE with some good explanation which does match opinions from one side of the camp in that StackOverflow debate. Yet there is the other camp, and as I see NPE usage is widespread and even officially sanctioned. Though I'd believe only in Effective Java and not in original Java docs, as there is enough of things done poorly in core Java to follow it everywhere.

PR #578, opened 2018-01-05
Change exception used

weidox:
NullPointerException is for null reference access, not for parameter validation, and in general should not be throwed by user code. Use some other exception, like IllegalArgumentException or IllegalStateException depending on circumstances.


weidox:
I signed CLA.

PR #587, opened 2018-01-22
Moe sync

ronshapiro:


ronshapiro:
@eamonnmcmanus looks like we got another `-proc:none` failure

PR #602, opened 2018-03-03
Updated AutoValue version from 1.5 to 1.5.3

tentenponce:


tentenponce:
I signed it!

eamonnmcmanus:
We've updated this ourselves, but it does draw attention to the fact that we need something better than a manual update process.

Issue #268, opened 2015-09-10
Separate AutoValue Annotations and Processor into different dependencies.


rharter:
It would be really helpful, especially in constrained environments like Android, to have the Annotations and the Processor separated into different dependencies.  The current structure leads to unnecessarily bloated output binaries.

Currently on Android, if the user includes AutoValue using `compile 'com.google.auto.value:auto-value:1.2'` then all of the processing classes and all of the shaded classes are included in the final binary.  There are ways around this, but they are hacky and unclear.

Separating the annotations from the processor would allow something like the following:

```
compile 'com.google.auto.value:auto-value-annotation:1.2'
apt `com.google.auto.value:auto-value:1.2
```

This naming would allow existing users to update smoothly (assuming `auto-value` depends on `auto-value-annotation`)

This would only include the minimal set of classes (annotations) required in the output binary and ensure that all other classes, including shaded classes, aren't unnecessarily included.

JakeWharton:
Best workaround I've found so far is using `apt` for the dependency and just copy/pasting the `AutoValue` annotation type into your app. But it would be great to not have to do this.

JakeWharton:
Also, +1 :grinning:

alangpierce:
AutoValue doesn't have any runtime dependencies (since the `AutoValue` annotation itself  only has `SOURCE` retention), so you can use the `provided` scope to avoid including anything in your binary, like this:

```
provided 'com.google.auto.value:auto-value:1.1'
apt 'com.google.auto.value:auto-value:1.1'
```

JakeWharton:
This is what everyone is doing. It's bad for two reasons:

* Obvious duplication across configurations is error prone and annoying
* It exposes the types and all dependencies on the classpath for IDE autocompletion

tbroyer:
While I tend to agree:

* Auto Value no longer depends on Velocity, so all dependencies are now shaded
* #264 would hide shaded dependencies from IDE autocompletion
* the processor itself could be similarly hidden from autocompletion if that's a problem (people using Maven would see the processor if not hidden that way, even when split into a separate JAR, as maven doesn't have first-class support for annotation processors – i.e. an "apt scope" or maven-compiler-plugin configuration).

[...]

Issue #290, opened 2015-11-24
AutoValue has SOURCE RetentionPolicy


hgschmie:
Is there a deeper meaning behind this? Right now, this makes it really hard for tooling to look at code and do special things if a class is autogenerated. (e.g. I want to dynamically discover the correct getters to reflect on a auto generated bean). 

As the use case: 

A public bean A has a `public abstract String foo()` method. The AutoValue_A bean also has a `public abstract String foo()` method, however the class itself is package private. 

When calling `object.getClass().getDeclaredMethods()` on an instance of `AutoValue_A`, there will be a method object representing the `foo()` method. However, calling invoke on that method with the instance as parameter will result in an `IllegalAccessException`. What is necessary is to find the super class (which is `A`) and use this class object to do getDeclaredMethods() and then call invoke on the method object which represents `foo()` in the `A` object.

So what I suggested to do is to traverse up the super class chain until the class having the `@AutoValue` annotation is encountered, then call `getDeclaredMethods()` on that class object.

`AutoValue` has SOURCE retention policy. Bummer. Is it possible to make that `RUNTIME`?

kevinb9n:
On Mon, Nov 23, 2015 at 4:33 PM, Henning Schmiedehausen <
notifications@github.com> wrote:

> Is there a deeper meaning behind this? Right now, this makes it really
> hard for tooling to look at code and do special things if a class is
> autogenerated.
>
One of AutoValue's core values is that no one should ever need to care
whether a type was implemented using AutoValue or some other means.


> A public bean A has a public abstract String foo() method. The
> AutoValue_A bean also has a public abstract String foo() method, however
> the class itself is package private.
>
> When calling object.getClass().getDeclaredMethods() on an instance of
> AutoValue_A, there will be a method object representing the foo() method.
> However, calling invoke on that method with the instance as parameter will
> result in an IllegalAccessException. What is necessary is to find the
> super class (which is A) and use this class object to do
> getDeclaredMethods() and then call invoke on the method object which
> represents foo() in the A object.
>
> So what I suggested to do is to traverse up the super class chain until
> the class having the @AutoValue annotation is encountered, then call
> getDeclaredMethods() on that class object.
>
Sounds like a general quirk with Java reflection. First question is, of
course, whether you can get away with setAccessible(true). But the other
generalized solution is to examine *all* supertypes that declare an
override-compatible signature, until you find a public one. It shouldn't
matter whether that's the @AutoValue class or not, should it?


-- 
[...]

PR #606, opened 2018-03-12
Moe Sync

ronshapiro:

Commits:
=====
<p> Create an auto-value-annotations artifact separate from the processor

Fixes https://github.com/google/auto/issues/268

I followed the instructions in https://maven.apache.org/guides/mini/guide-using-one-source-directory.html because it seemed like the easiest way forward without moving around any files.

RELNOTES=`@AutoValue`, `@AutoAnnotation`, `@AutoOneOf`, and `@Memoized` are now in a separate artifact, `auto-value-annotations`. This allows users to specify the annotations in compile scope and the processor in an annotation processing scope, without leaking the processor to a release binary. To upgrade to this version of auto-value, you'll need to add this new artifact as a dependency.

eb7b4ae7847859a2cac68a3a0e4d1ff6b25ccade

### DIFF
--- a/value/userguide/index.md
+++ b/value/userguide/index.md
@@ -76,19 +76,29 @@ Maven users should add the following to the project's `pom.xml` file:
 ```xml
 <dependency>
   <groupId>com.google.auto.value</groupId>
+  <artifactId>auto-value-annotations</artifactId>
+  <version>1.6</version>
+</dependency>
+<dependency>
+  <groupId>com.google.auto.value</groupId>
   <artifactId>auto-value</artifactId>
-  <version>1.5</version>
+  <version>1.6</version>
   <scope>provided</scope>
 </dependency>
 ```
+
+Alternatively, instead of using the `provided` scope, you can add the second
+dependency to the
+[`annotationProcessorPaths`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths)
+section.
 
 Gradle users should install the annotation processing plugin [as described in
 these instructions][tbroyer-apt] and then use it in the `build.gradle` script:
 
 ```groovy
 dependencies {
-  compileOnly "com.google.auto.value:auto-value:1.5"
-  apt         "com.google.auto.value:auto-value:1.5"
+  compileOnly "com.google.auto.value:auto-value-annotations:1.6"
+  apt         "com.google.auto.value:auto-value:1.6"
 }
 ```
 
--- a/value/annotations/pom.xml
+++ b/value/annotations/pom.xml
@@ -25,13 +25,12 @@
   </parent>
 
   <groupId>com.google.auto.value</groupId>
-  <artifactId>auto-value-pom-aggregator</artifactId>
+  <artifactId>auto-value-annotations</artifactId>
   <version>HEAD-SNAPSHOT</version>
-  <name>AutoValue</name>
+  <name>AutoValue Annotations</name>
   <description>
     Immutable value-type code generation for Java 1.6+.
   </description>
-  <packaging>pom</packaging>
 
   <scm>
     <url>http://github.com/google/auto</url>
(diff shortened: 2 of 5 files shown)


value/userguide/index.md
```diff
@@ -74,21 +74,31 @@ examples short and simple.
 Maven users should add the following to the project's `pom.xml` file:
 
 ```xml
+<dependency>
+  <groupId>com.google.auto.value</groupId>
+  <artifactId>auto-value-annotations</artifactId>
```
adriancole: unless the annotations are now runtime retention... is it really necessary to add this as a strict dep to maven projects?

eamonnmcmanus: The annotations (at least most of them) now have `CLASS` retention, so it isn't completely pointless to have them at runtime. I suppose we could tell people that they can give this dependency `provided` scope if they want, but is it worthwhile for such a tiny dependency?

cpovirk: FYI: Guava (and, I recently learned from jbduncan, [JUnit 5](https://github.com/junit-team/junit5/issues/1105)) have stopped using `provided` scope for annotation dependencies after [various](https://github.com/google/guava/issues/1018) [discussions](https://github.com/google/guava/issues/2721) [of](https://github.com/google/guava/issues/2824) [problems](https://github.com/google/guava/issues/2837). In short, `provided` [doesn't mean "compile-time"](https://issues.apache.org/jira/browse/MNG-2205?focusedCommentId=16267311&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16267311) as you'd probably hope; instead, it means "compile-time for this project and missing entirely in downstream compilations." This causes problems for people who enable all `javac` warnings or use some annotation processors.

eamonnmcmanus: We should probably add a note in `index.md` suggesting that people *not* use `provided` scope, then, with your [link](https://issues.apache.org/jira/browse/MNG-2205?focusedCommentId=16267311&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16267311).

tbroyer: Hmm, then the Gradle snippet should probably use `api` (for [`java-library`](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph) and [Android](https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#new_configurations) projects) or `compile` (for all others).

Though this actually depends whether your autovalues are part of your public API or not.

BTW, `apt` should be updated to `annotationProcessor`, and with Gradle 4.6 the `net.ltgt.apt` plugin is no longer required (though it still provides useful features).

So how about:

> Gradle users should use the following in their `build.gradle` script:
> 
> ```groovy
> dependencies {
>   // Use 'api' rather than 'compile' for Android or java-library projects.
>   compile             "com.google.auto.value:auto-value-annotations:1.6"
>   annotationProcessor "com.google.auto.value:auto-value:1.6"
> }
> ```
> You may want, or have to, install the [`net.ltgt.apt`][tbroyer-apt] plugin in non-Android projects.
> 
> [tbroyer-apt]: https://github.com/tbroyer/gradle-apt-plugin

Note that I updated the link to my plugin to point to the GitHub repo where the README provides documentation, including about companion plugins for better IDE support, whereas the Gradle Plugin Portal only has code snippets (I've had a couple reports already from people who hadn't seen the companion plugins, coming from the Plugin Portal).

And in the note you suggest to add about `<scope>provided</scope>` (or `<optional>true</optional>`), add a note about the `compileOnly` and `implementation` Gradle configurations (that's the technical name in Gradle, "configurations", not "scopes").

artem-zinnatullin: Hm, shouldn't be `compileOnly` recommended for Gradle users then? 

This seems like best configuration: annotations will be available to javac and annotation processors (including 3rd-party "extension" processors) of the module, but won't be included into jar nor will be transitively available to other modules.

Afaik philosophy of Auto* was to be hidden from runtime introspection by default https://github.com/google/auto/issues/290#issuecomment-328547302

tbroyer: @artem-zinnatullin See comments above, specifically https://github.com/google/auto/pull/606#discussion_r177278685 and https://github.com/google/auto/pull/606#discussion_r177441884. The issue is about being present at compile-time (or static analysis time) of _downstream_ projects.

PR #619, opened 2018-03-28
Moe Sync

ronshapiro:

Commits:
=====
<p> Update AutoValue's Gradle documentation to account for the built-in annotation processor configuration.

The standalone apt plugin is now only required for use on previous versions.

07426cba781efdd9185a5e9553694fcce201bbbd

### DIFF
--- a/value/userguide/index.md
+++ b/value/userguide/index.md
@@ -92,15 +92,19 @@ dependency to the
 [`annotationProcessorPaths`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths)
 section.
 
-Gradle users should install the annotation processing plugin [as described in
-these instructions][tbroyer-apt] and then use it in the `build.gradle` script:
+Gradle users can declare the dependencies in their `build.gradle` script:
 
 ```groovy
 dependencies {
-  compileOnly "com.google.auto.value:auto-value-annotations:1.6"
-  apt         "com.google.auto.value:auto-value:1.6"
+  // Use 'api' rather than 'compile' for Android or java-library projects.
+  compile             "com.google.auto.value:auto-value-annotations:1.6"
+  annotationProcessor "com.google.auto.value:auto-value:1.6"
 }
 ```
+
+Note: If you are using a version of Gradle prior to 4.6, you must apply an
+annotation processing plugin [as described in these instructions][tbroyer-apt]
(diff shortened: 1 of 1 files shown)


value/userguide/index.md
```diff
@@ -92,16 +92,20 @@ dependency to the
...
-  apt         "com.google.auto.value:auto-value:1.6"
+  // Use 'api' rather than 'compile' for Android or java-library projects.
+  compile             "com.google.auto.value:auto-value-annotations:1.6"
+  annotationProcessor "com.google.auto.value:auto-value:1.6"
 }
 ```
 
+Note: If you are using a version of Gradle prior to 4.6, you must apply an
+annotation processing plugin [as described in these instructions][tbroyer-apt]
+and use the `apt` configuration instead of `annotationProcessor`.
```
tbroyer: No need to use `apt`; latest version of the plugin uses `annotationProcessor`

value/userguide/index.md
```diff
@@ -92,16 +92,20 @@ dependency to the
...
+
 [tbroyer-apt]: https://plugins.gradle.org/plugin/net.ltgt.apt
```
tbroyer: Link to GitHub to get the documentation? (and not just the snippet to apply it)

value/userguide/index.md
eamonnmcmanus: Thanks, we'll delete the business of using the `apt` configuration.

value/userguide/index.md
eamonnmcmanus: There's a GitHub link right at the top of that page. Is there a link you'd rather see instead of the one here?

tbroyer: I was referring to https://github.com/tbroyer/gradle-apt-plugin, or https://github.com/tbroyer/gradle-apt-plugin/blob/master/README.md (see https://github.com/google/auto/pull/606#discussion_r177486587)

That link doesn't itself include instructions (it defers to https://plugins.gradle.org/plugin/net.ltgt.apt) but has the documentation for the plugin, and talks about setting up your IDE, which is facilitated using other plugins. Both links reference the other, so use the one you prefer, but I've had issues reported about not seeing generated code in IDEs from people that came from the Gradle Plugin Portal and didn't click through to the README (they did find where to report the issue though, so it might be that they just didn't bother looking things up before creating an issue).

I don't have a strong opinion on that one.

PR #623, opened 2018-04-04
@AutoValue generated class removes the leading space from final or abstract ketword

### LINKED ISSUE #622: @AutoValue generated class has extra leading space .
@AutoValue generated class has extra leading space with final or abstract keyword.

Sathawale27:
Closes #622


eamonnmcmanus:
Thanks! Our complicated mirroring system means that we won't be merging this PR directly, but an equivalent and duly-credited PR should be merged shortly.

PR #353, opened 2016-07-01
Add support for Nullable on Optional property builders

kenzierocks:
This allows an `@Nullable` annotation to exist on a `Builder` property setter for an `Optional` property, like so: `Builder bar(@Nullable String bar);`. It uses `ofNullable`/`fromNullable` to convert the value to the correct `Optional` instance. ~~This PR also replaces the logic in the `Builder` property getters to use `ofNullable`/`fromNullable` instead of an `if ($p == null)` check.~~ (Edit: This doesn't work with `Optional` primitives + wrapper classes.)

This solves my request here: https://github.com/google/auto/issues/278#issuecomment-226660312, and #384.

PR #634, opened 2018-05-07
Moe Sync

ronshapiro:

Commits:
=====
<p> Allow an Optional property to be set in a builder through a method with a @Nullable parameter.

This is based on https://github.com/google/auto/pull/353 by Kenzie Togami.

Closes https://github.com/google/auto/pull/353.

RELNOTES:
  Allow an Optional property to be set in a builder through a method with a @Nullable parameter.
  NOTE: As a side-effect of this change, a setter for a @Nullable property must have its parameter also marked @Nullable. Previously, if the @Nullable was not a TYPE_USE @Nullable, it was copied from the getter to the parameter of the generated setter implementation. For TYPE_USE @Nullable it was already necessary to mark the setter parameter @Nullable.

53e8144d4a45e3c2b8664fa3c041bf2e62d77620

-------

<p> Replace http://errorprone.info with https://errorprone.info since that site
serves from https now

13bde05677d369d85caf134d9d80c43697d4d193


smac89:
@kenzierocks please sign :raised_hands:

PR #636, opened 2018-05-09
 hashCode() will take global variable while generating method

### LINKED ISSUE #635: hashCode() won't work properly in @AutoValue generated class
hashCode() will not work properly, if there is an any type of `abstract h$();` method in @AutoValue annotated class. While generating hashCode() method, It will take h$ local parameter instead of global one.

Sathawale27:
Closes #635


Sathawale27:
Oops! 
Forgot to make changes into tests. will do it shortly.

eamonnmcmanus:
Thanks for the PR! But I think we don't want to make this change, for the reasons explained in #635.

PR #639, opened 2018-05-16
mark scope as provided

honnix:
Is this supposed to be `provided` because it is only needed at compile time?

### DIFF
--- a/service/README.md
+++ b/service/README.md
@@ -47,7 +47,7 @@ In order to activate metadata generation you will need to include
 `auto-service-${version}.jar` in your build at compile time.
 
 In a Maven project, one would include the `auto-service` 
-artifact as an "optional" dependency:
+artifact as a "provided" dependency:
 
 ```xml
 <dependencies>
@@ -55,7 +55,7 @@ artifact as an "optional" dependency:
     <groupId>com.google.auto.service</groupId>
     <artifactId>auto-service</artifactId>
     <version>${version}</version>
-    <optional>true</optional>
+    <scope>provided</scope>
   </dependency>
 </dependencies>
 ```


honnix:
Or can be both if really needed.

honnix:
I signed it!

davideicardi:
Any command from a maintainer on this? We should include `com.google.auto.service` in our jar or not?

tbroyer:
Wrt optional vs provided, it looks like the Maven community is divided on the subject (http://maven.40175.n5.nabble.com/lt-scope-gt-provided-lt-scope-gt-vs-lt-optional-gt-true-lt-optional-gt-td2852244.html; note the date!)

To answer your question @davideicardi, no, you don't need AutoService at runtime, the annotation has _source retention_ anyway so you couldn't even detect it on your service through reflection.

What the README should say however is something like (copy/pasted from Dagger, added optional back; feel free to use scope=provided if that works better for you, but it's now only a single annotation so you could also have it included transitively):
```xml
<dependencies>
  <dependency>
    <groupId>com.google.auto.service</groupId>
    <artifactId>auto-service-annotations</artifactId>
    <version>${version}</version>
    <optional>true</optional>
  </dependency>
</dependencies>
<build>
  <plugins>
    <plugin>
      <groupId>org.apache.maven.plugins</groupId>
      <artifactId>maven-compiler-plugin</artifactId>
      <version>3.6.1</version>
      <configuration>
        <annotationProcessorPaths>
          <path>
            <groupId>com.google.auto.service</groupId>
            <artifactId>auto-service</artifactId>
            <version>${version}</version>
          </path>
        </annotationProcessorPaths>
      </configuration>
    </plugin>
  </plugins>
</build>
```

nglorioso:
https://github.com/google/auto/commit/7dc5450ec1abdd6473a5dd11dcb19f4115ff7a89#diff-8c9513528e170b62556353ac79283c4f documents the strategy described by tbroyer@

PR #643, opened 2018-06-18
Adds support for Factory depending on another Factory (fixes #297)

### LINKED ISSUE #297: AutoFactory: Composition of factories broken
When referencing a auto generated factory in another auto generated factory the reference is not fully qualified and thus can not be resolved. Here is an example:

``` java
package autoFactoryTest.sub;

@AutoFactory
public class B {
}
```

``` java
package autoFactoryTest;

import autoFactoryTest.sub.BFactory;

@AutoFactory
public class A {
    public A(BFactory b) {
    }
}
```

The reference to `BFactory` from `AFactory` is not fully qualified which cannot be resolved:

``` java
import javax.annotation.Generated;
import javax.inject.Inject;
@Generated("com.google.auto.factory.processor.AutoFactoryProcessor")
public final class AFactory {
  @Inject
  public AFactory() {
  }
  public A create(BFactory b) {
    return new A(b);
  }
}
```

compilation error:

```
[ERROR] diagnostic: target/generated-sources/apt/autoFactoryTest/AFactory.java:8: error: cannot find symbol
```

winterDroid:
This will address the issue that the composition of factories is broken.
Builds a map of factory name to to be generated factory which will be used to guess the type
Current workaround is to specify the fully qualified class name

Open Issue:
If two factories will have the same name it can not be guaranteed that the proper class will be selected

### DIFF
--- a/factory/src/main/java/com/google/auto/factory/processor/AutoFactoryProcessor.java
+++ b/factory/src/main/java/com/google/auto/factory/processor/AutoFactoryProcessor.java
@@ -23,11 +23,15 @@ import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
 import com.google.googlejavaformat.java.filer.FormattingFiler;
+import com.squareup.javapoet.ClassName;
+import com.squareup.javapoet.TypeName;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


winterDroid:
I also thought about a more "sophisticated" way:
If a `@Provided` parameter is supposed to be a factory, the element the factory is used for to generate, should be passed as an argument to the annotation. In my example it would be like:
```
@AutoFactory
final class SimpleClassDependingOnFactory {
  private final PublicClassFactory depA;

  SimpleClassDependingOnFactory(@Provided(PublicClass.class) PublicClassFactory depA) {
    this.depA = depA;
  }
}
```
In this way the to be generated factory can be reliably retrieved. Besides that I was wondering whether the build should fail if there is a parameter of type `ErrorType`?

netdpb:
I think what's going on is that `BFactory` hasn't been generated yet, and some part of AutoFactory (or JavaPoet) is not understanding the name of an `ErrorType`. If `AutoFactoryProcessor` extended `BasicAnnotationProcessor`, it could defer processing of `A` until a later round.

I don't know what to do if there are circular references—if A needs a BFactory and B needs an AFactory.

If I'm right, then there's a way forward that isn't using heuristics.

winterDroid:
Yes that's what I thought as well. Using the `BasicAnnotationProcessor` is a great idea. I can look into it and check how to tackle the circular reference issue

PR #647, opened 2018-06-27
avoid java.beans.Introspector import

buchgr:
Starting with JDK9, the `java.beans` package is defined inside
the `java.desktop` module which is ~10MiB in size. Applications
wanting to run on a minimized JVM (via jlink) should not need
to pull in 10MiB to use Auto.

------

I work on Bazel (https://bazel.build) and we are happy users of Auto. A Bazel binary ships with an embedded JDK and we are currently working on reducing our binary size. We can run with a JDK as small as 19 MiB, but the jdk.desktop module blows this up to 29 MiB. It would be great if we didn't have to depend on it. Thanks!

cc: @meisterT

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/PropertyNames.java
+++ b/value/src/main/java/com/google/auto/value/processor/PropertyNames.java
@@ -1,0 +1,31 @@
+package com.google.auto.value.processor;
+
+/**
+ * Helper methods to create property names.
+ */
+final class PropertyNames {
+
+  /**
+   * Like {@link #decapitalize(String)} but returns a property name that
+   * starts with two capital letters as is (does nothing).
+   */
+  public static String decapitalizeIgnoreAcronyms(String propertyName) {
+    if (propertyName != null &&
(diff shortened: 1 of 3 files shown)


value/src/main/java/com/google/auto/value/processor/PropertyNames.java
```diff
@@ -0,0 +1,31 @@
...
+        Character.isUpperCase(propertyName.charAt(1))) {
+      return propertyName;
+    }
+    return decapitalize(propertyName);
+  }
+
+  /**
+   * Returns the {@code propertyName} with its first character in lower case.
+   */
+  public static String decapitalize(String propertyName) {
```
eamonnmcmanus: I think you also meant to delete the original version of this method in BuilderMethodClassifier?

eamonnmcmanus:
Thanks! I'll convert this into an internal change and send it to you for review.

buchgr:
Merged as 8786426a51f4d1a9548bdc9a9a3594cb786102d

Issue #615, opened 2018-03-25
Support incremental annotation processing 


stephanenicolas:
Gradle 4.7 provides support for incremental annotation processing.
https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing

It would be nice to see auto (at least auto-value) supporting it. The change is not that big as auto value is an isolating AP, and even simpler as it doesn't handle inheritance.

eamonnmcmanus:
We'd be happy to receive a PR from someone who has tested that this works. Just a couple of cautionary notes:

* The cited document says that processors "can't read resources". AutoValueProcessor does read resources, namely the template files it uses to generate code. I suspect that the document means that a processor can't read *arbitrary* resources, rather than these fixed ones.

* AutoValue *does* handle inheritance. For example the `@AutoValue` class can inherit abstract methods from a superclass or superinterface and those will be implemented. But I think we are still within the documented constraint "Must make all decisions about an element based on information reachable from its AST." The only exception I can think of is that we use `javax.annotation.Generated` or `javax.annotation.processing.Generated` depending on what is available. That's very unlikely to matter in practice.

stephanenicolas:
Hi @eamonnmcmanus , 

you're right
1. this resource doesn't matter. All is about reproducibility of the processing. So a fixed resource is fine. It would be more about reading stuff that can change (updated data, time, stuff from internet..)
2. I didn't know that, cool then. Does auto value support inheritance across different modules already ? I mean if a subclass can inherit a super class in a different module, does all work well (and efficiently? )

They are different approaches to resolve the second point. A naive approach is simply to crawl up the hierarchy again and re process all annotations in super classes. That would work wonderfully here. 
Some different approoaches try to optimize this process (for instance the generated code for a subclass can call the generated code from a super class) and they are harder to make compatible with modularization or incremental annotation processing.

eamonnmcmanus:
The supertype processing is just used to determine what methods are inherited by the `@AutoValue` class. We don't allow one `@AutoValue` class to inherit from another, and even if we did you still couldn't inherit from (or otherwise reference) a generated `AutoValue_Foo` class in another module, so I don't think there are any concerns about modularity. As far as I can tell, everything should Just Work if we put the right thing in `META-INF/gradle/incremental.annotation.processors`.

stephanenicolas:
[...]

PR #657, opened 2018-07-31
Make auto-service and auto-value incremental processors (for Gradle)

tbroyer:
#615

### DIFF
--- a/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
+++ b/value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
@@ -109,6 +109,38 @@ public abstract class AutoValueExtension {
   }
 
   /**
+   * Indicates to an annotation processor environment supporting incremental annotation processing
+   * (currently a feature specific to Gradle starting with version 4.8) the incremental type of an
+   * Extension.
+   *
+   * <p>The constants for this enum are ordered by increasing performance (but also constraints).
+   *
+   * @see <a
+   *     href="https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing">Gradle
+   *     documentation of its incremental annotation processing</a>
+   */
+  public enum IncrementalExtensionType {
+    UNKNOWN,
+    AGGREGATING,
+    ISOLATING
+  }
+
+  /**
+   * Determines the incremental type of this Extension.
+   *
+   * <p>The {@link ProcessingEnvironment} can be used, among other things, to obtain the processor
+   * options, using {@link ProcessingEnvironment#getOptions()}.
+   *
+   * <p>The actual incremental type of the AutoValue processor as a whole will be the loosest
+   * incremental types of the Extensions present in the annotation processor path. The default
+   * returned value is {@link IncrementalExtensionType#UNKNOWN}, which will disable incremental
+   * annotation processing entirely.
+   */
+  public IncrementalExtensionType incrementalType(ProcessingEnvironment processingEnvironment) {
+    return IncrementalExtensionType.UNKNOWN;
+  }
+
+  /**
    * Determines whether this Extension applies to the given context.
    *
    * @param context The Context of the code generation for this class.
--- a/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
+++ b/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
@@ -33,6 +33,7 @@ import com.google.common.collect.ImmutableSet;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -43,7 +44,6 @@ import java.util.Set;
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.annotation.processing.Processor;
 import javax.annotation.processing.SupportedAnnotationTypes;
-import javax.annotation.processing.SupportedOptions;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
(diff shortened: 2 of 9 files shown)


service/pom.xml
```diff
@@ -40,6 +40,12 @@
   </scm>
 
   <dependencies>
+    <dependency>
```
oehme: I don't think the convenience library will work, because this build uses `proc:none`. You'll have to do the registration the manual way (which isn't that much harder).

tbroyer: Ha! good catch; fixed.

tbroyer:
@ronshapiro Is this something you'd like to see me pursuing? (particularly wrt the API additions for autovalue extensions)

ronshapiro:
I'll let @eamonnmcmanus comment on AutoValue. AutoService support sounds fine, though I imagine it's often not in large builds?

oehme:
I agree AutoService is rare in large projects. But since it's so easy to make incremental, there's really no reason not to :)

bingranl:
Thanks for your PR! @tbroyer I am from Google android build team. I am wondering if there is a good way to test it which can also be used for other APs. Do you have something on your end? @oehme (I know there is a gradle tooling api)

oehme:
What do you have in mind beyond the tooling API? The test cases would have to be processor specific.

bingranl:
What is the reason that tests have to be specific? Is it because you have to write functional/unit test to make sure existing functionality works? I am hoping to find a way to tell if it is incremental or not as Gradle doc mentions  "As a user you can see which annotation processors are triggering full recompilations in the --info log"

oehme:
I don't think it's worth testing whether you successfully registered, that's trivial and unlikely to break. The interesting part is testing whether your processor actually honors the contract (i.e. provides originating elements, doesn't use any hidden channels like java.io etc.). These kinds of tests (if you want to add them) would be processor specific.

bingranl:
Then what can --info log tell us? You mean even if it shows this AP is incremental, it is very likely this AP breaks some contracts and are not actually incremental?

oehme:
My point is that a bad processor may do things that Gradle cannot detect. So we'll run incremental, but possibly fail. That's why it's important that the processor author reads and understands the restrictions in the documentation and makes sure their processor follows these restrictions. This can be done with code review and supported by testing where necessary.

auto-value and auto-service are clearly well-behaved, they only use the annotation processing APIs as intended. But I've seen other processors that access compiler internals or use `java.io`. Those need to change their approach before they can declare themselves as incremental.

eamonnmcmanus:
What's the status of this PR? Is it still Work In Progress as the title says, or could we consider taking it as it stands?

value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
```diff
@@ -112,6 +116,23 @@ public synchronized void init(ProcessingEnvironment processingEnv) {
...
+    AutoValueExtension.IncrementalExtensionType incrementalType = extensions.stream()
+        .map(e -> e.incrementalType(processingEnv))
+        .min(Comparator.naturalOrder())
+        .orElse(AutoValueExtension.IncrementalExtensionType.ISOLATING);
+    switch (incrementalType) {
+      case ISOLATING:
+        return ImmutableSet.of(OMIT_IDENTIFIERS_OPTION, IncrementalAnnotationProcessorType.ISOLATING.getProcessorOption());
+      case AGGREGATING:
+        return ImmutableSet.of(OMIT_IDENTIFIERS_OPTION, IncrementalAnnotationProcessorType.AGGREGATING.getProcessorOption());
+      case UNKNOWN:
```
eamonnmcmanus: I would delete this, or delete the default case and follow the switch with `throw new AssertionError(incrementalType)`.

value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
```diff
@@ -84,6 +84,18 @@
     Set<ExecutableElement> abstractMethods();
   }
 
+  // TODO: javadoc
```
eamonnmcmanus: Can you add this? Don't forget to mention that the order is important.

tbroyer:
> What's the status of this PR? Is it still Work In Progress as the title says, or could we consider taking it as it stands?

You may want to add tests, and/or an option (or system property) to override / disable incremental processing as a whole (like it's being done for Dagger).

I've otherwise addressed your feedback.

value/src/main/java/com/google/auto/value/extension/AutoValueExtension.java
```diff
@@ -108,6 +108,38 @@
...
+   * Extension.
+   *
+   * <p>The constants for this enum are ordered by increasing performance (but also constraints).
+   *
+   * @see <a
+   *     href="https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing">Gradle
+   *     documentation of its incremental annotation processing</a>
+   */
+  public enum IncrementalExtensionType {
+    UNKNOWN,
```
eamonnmcmanus: I feel like the individual enum constants should have an understandable summary of what they mean. Here's what I've written in the copy of this PR that I'm importing into Google's private repo:

```java
    /**
     * The incrementality of this extension is unknown, or it is neither aggregating nor isolating.
     */
    UNKNOWN,

    /**
     * This extension is <i>aggregating</i>, meaning that it may generate outputs based on several
     * annotated input classes and it respects the constraints imposed on aggregating processors.
     * It is unusual for AutoValue extensions to be aggregating.
     *
     * @see <
     *     href="https://docs.gradle.org/current/userguide/java_plugin.html#aggregating_annotation_processors">Gradle
     *     definition of aggregating processors</a>
     */
    AGGREGATING,

    /**
     * This extension is <i>isolating</i>, meaning roughly that its output depends on the
     * {@code @AutoValue} class and its dependencies, but not on other {@code @AutoValue} classes
     * that might be compiled at the same time. The constraints that an isolating extension must
     * respect are the same as those that Gradle imposes on an isolating annotation processor.
     *
     * @see <
     *     href="https://docs.gradle.org/current/userguide/java_plugin.html#isolating_annotation_processors">Gradle
     *     definition of aggregating processors</a>
     */
    ISOLATING
```

Does that seem right? I'm actually finding it hard to imagine what an aggregating extension might be trying to do.

tbroyer: > Does that seem right?

SGreatTM :wink:

> I'm actually finding it hard to imagine what an aggregating extension might be trying to do.

IIUC, looking up another class by name, e.g using a naming rule and/or an annotation property value (or processor option).
Probably unusual, but not impossible… :man_shrugging:

Issue #670, opened 2018-10-11
AutoValue generated code failing new Error Prone UnnecessaryParentheses check


msridhar:
Error Prone 2.3.2 was just released and includes a new [UnnecessaryParentheses check](http://errorprone.info/bugpattern/UnnecessaryParentheses).  We compile with `-Werror` and noticed that AutoValue-generated code fails this check.  Here's an example getting flagged (I added the `// warning here` comments):

```java
  @Override
  public boolean equals(Object o) {
    if (o == this) {
      return true;
    }
    if (o instanceof DataFlow.AnalysisParams) {
      DataFlow.AnalysisParams that = (DataFlow.AnalysisParams) o;
      return (this.transferFunction.equals(that.transferFunction())) // warning here
          && (this.cfg.equals(that.cfg())); // warning here
    }
    return false;
  }
```

For now we've worked around this by turning off Error Prone for generated code, and we may actually stick with that in the long term.  So this is more of an FYI than a significant problem.

eamonnmcmanus:
Thanks for the report! I agree that it is probably not worth running Error Prone on generated code. This particular issue is easy enough to fix though so we'll probably do that.

PR #671, opened 2018-10-22
Moe Sync

ronshapiro:

Commits:
=====
<p> Remove unnecessary parentheses from the generated equals(Object) method.

Fixes https://github.com/google/auto/issues/670.

RELNOTES=Remove unnecessary parentheses from the generated equals(Object) method in @AutoValue classes.

46e0bf4c68f227136ec689a72f084cd0a4d4802b

-------

<p> Have MemoizeExtension recognise @Nullable as a type annotation.

It currently only recognises @Nullable as a method annotation, typically for javax.annotation.Nullable, but does not recognise the TYPE_USE version from the checker framework.

We also annotate the corresponding constructor parameter, though currently only for the TYPE_USE case.

While testing this I discovered that we can pick up two versions of MemoizeExtension because it changed packages. Specifically I saw Maven picking up AutoValue 1.5.3 via compile-testing. The way ServiceLoader works means that we will load and run both versions of MemoizeExtension in this case. So I added a hack to AutoValueProcessor so it will ignore the old version if it sees it.

RELNOTES=MemoizeExtension recognises @Nullable type annotations, not just method annotations.

8e13fa94c7cf9e6b5f99029c37f02d69f139585c

### DIFF
--- a/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java
+++ b/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java
@@ -62,6 +62,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.TypeParameterElement;
+import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Elements;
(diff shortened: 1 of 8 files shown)


ronshapiro:
Travis seemed to have missed this PR during Github's partial outage yesterday. Closing and reopening with the hopes that it will run the tests.

Issue #683, opened 2018-11-28
Uplift guava version to 24.1.1+ to fix CVE-2018-10237


venki09:
https://nvd.nist.gov/vuln/detail/CVE-2018-10237 affects autovalue as it is using guava version 23 currently. I think this needs to be updated to fix the vulnerability.

tbroyer:
Guava is only used by the processors, it's not required at runtime, so Auto projects aren't impacted by this Guava vulnerability.

venki09:
maven dependency-check-plugin is flagging it as a vulnerability though. Is that a false positive then?

eamonnmcmanus:
It is indeed a false positive. But there's no reason we couldn't upgrade to a more recent Guava anyway.

Issue #686, opened 2018-12-01
AutoValue does not find Generated annotation on Java 11 with -source 1.8


PhilippWendler:
We use AutoValue in a Java 8 project and want to test compilation on Java 11. Unfortunately, in this scenario AutoValue does not find a `Generated` annotation on the class path, and thus skips adding it. However, we need this annotation to be present because otherwise error-prone warns about the generated source code (unnecessary parentheses etc.).

For reproduction take the following source:
```java
import com.google.auto.value.AutoValue;

@AutoValue
public abstract class Test {

  public static Test create() {
    return new AutoValue_Test();
  }
}
```

It can be compiled using
```
javac -processorpath auto-value.jar -cp auto-value-annotations.jar -source 1.8 *.java
```
On Java 8, 9, and 10, this produces
```java
...
import javax.annotation.Generated;
@Generated("com.google.auto.value.processor.AutoValueProcessor")
...
```

On Java 11 this produces
```java
...
// Generated by com.google.auto.value.processor.AutoValueProcessor
...
```
Bumping `-source` to `1.9` produces the expected
```java
...
import javax.annotation.processing.Generated;
@Generated("com.google.auto.value.processor.AutoValueProcessor")
...
```

I am using the `openjdk:11-jdk-slim` Docker image, which has
```
$ java -version
openjdk version "11.0.1" 2018-10-16
OpenJDK Runtime Environment (build 11.0.1+13-Debian-2)
OpenJDK 64-Bit Server VM (build 11.0.1+13-Debian-2, mixed mode, sharing)
$ javac -version
javac 11.0.1
```

I am actually not sure if this is a problem in AutoValue itself or if it is a general problem with Java 11. Particularly I do not understand why it fails only with Java 11, given that it works on Java 10.

But nevertheless we would be interested in knowing how to make the combination Java 8 compatible source + AutoValue + Java 11 + Error-Prone work, or any other suggestions on how to proceed.

tbroyer:
Java 11 no longer includes `javax.annotation.Generated`, so it's no wonder that javac doesn't find it; unless you include some dependency that provides it.

You could also possibly try with `--release 8` instead of `-source 1.8`, though I'm not sure it would change the outcome (without `javax.annotation.Generated` in the classmate)

PhilippWendler:
Thank you very much for your response. I was wondering why AutoValue wouldn't find the new `javax.annotation.processing.Generated` either, which should be present on Java 11.

[...]

PR #687, opened 2018-12-02
Moe Sync

ronshapiro:

Commits:
=====
<p> Ignore ECJ test in JDK 10 and 11 since ECJ doesn't work for us in those versions of the JDK.

b3d49ff80772b508a4103fa85098c3de6c61fd6f

-------

<p> Update to auto-parent 7 to get the latest version of Guava

Along the way, also clean up some explicit versions that are set in auto-parent

Fixes https://github.com/google/auto/issues/683

ce333c488246cecd0c49c66a664c475e3ed0b0f5

### DIFF
--- a/common/src/test/java/com/google/auto/common/OverridesTest.java
+++ b/common/src/test/java/com/google/auto/common/OverridesTest.java
@@ -75,12 +75,14 @@ import org.junit.runners.model.Statement;
  */
 @RunWith(Parameterized.class)
 public class OverridesTest {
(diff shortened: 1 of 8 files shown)


tbroyer:
Latest ECJ apparently is:
```xml
<dependency>
  <groupId>org.eclipse.jdt</groupId>
  <artifactId>org.eclipse.jdt.compiler.tool</artifactId>
  <version>1.2.300</version>
</dependency>
```
Though note that #686 applies to `com.google.auto.common.GeneratedAnnotationsTest#source8` so the test will fail with JDK 11. Using `--release 8` in this test makes it pass.

PR #689, opened 2018-12-15
Automatic-Module-Name added to all manifests.

TheIndifferent:
As described by http://branchandbound.net/blog/java/2017/12/automatic-module-name/ , adding `Automatic-Module-Name` allows us to move to Java 11 without worrying that the jar file name will change.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -85,6 +85,13 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>com.google.auto.common</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
+        </configuration>
       </plugin>
     </plugins>
   </build>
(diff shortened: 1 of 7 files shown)


tbroyer:
According to https://blog.joda.org/2018/03/jpms-negative-benefits.html you shouldn't rush adding a module name: it needs testing to make sure the module is "well formed" and actually "compatible" with JPMS.

BTW, absence of module name doesn't prevent you to move to Java 11, only to use the module path.

TheIndifferent:
Good point. I experiment with that more intensively first.

TheIndifferent:
Reopening to use this PR as a discussion.

From what I've learned and see so far, Auto seems to be pretty well structured and modules will be well-formed. Each module can use root package as its name, and that will be the package it will export, and they don't seem to conflict in any way.

The problematic part here is that `value` uses single source root and then compiler configuration is responsible for grouping it into separate artifacts. The `module-info.java` file has to be on the top of the source code, which means that one of two things has to happen:
1. additional source root has to be added to every project.
2. current source root has to split into classically structured maven projects.

Considering all of the above, and considering that `auto` are leaf dependencies, this PR is fast and easy way to allow others to more forward and use the module names in future-proof way, and then some more complex approach of actually creating `module-info.java` and changing the build procedure can be implemented. Of course this step can be skipped if you are ok with any of the 2 points mentioned above, to add `module-info.java` now.

Issue #679, opened 2018-11-16
AutoFactoryProcessor is not incremental in Gradle


davidburstromspotify:
As pointed out here
https://github.com/google/auto/issues/615#issuecomment-433682099
`AutoFactoryProcessor` is not incremental in `com.google.auto.factory:auto-factory:1.0-beta6`

nglorioso:
Fixed by #615 now

PR #708, opened 2019-03-06
Autofactory: Support Gradle incremental annotation processing

rciovati:
Changes in this PR:
- Make the `AutoFactoryProcessor` [Isolating](https://docs.gradle.org/5.2/userguide/java_plugin.html#isolating_annotation_processors). This is done using the same library/plugin used in AutoValue.
- Propagate the originating element down to the `Filer`. Isolating processors require you to provide the originating element for each file generated with the Filer API. The change makes sure every generate file has an associated originating element, which is the target class annotated with `@AutoFactory` or the enclosing class, in case there is one or more methods/constructors annotated with `@AutoFactory`.

Should fix https://github.com/google/auto/issues/679
Relates to https://github.com/gradle/gradle/issues/5277

### DIFF
--- a/factory/pom.xml
+++ b/factory/pom.xml
@@ -77,6 +77,7 @@
       <groupId>net.ltgt.gradle.incap</groupId>
       <artifactId>incap</artifactId>
       <version>0.2</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>net.ltgt.gradle.incap</groupId>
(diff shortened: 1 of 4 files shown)


rciovati:
Is this PR something you might be interested in or should I just close it? cc @ronshapiro

ronshapiro:
I've had the change sitting and I totally forgot about it. I had a slightly simpler approach. Should be out shortly.

rciovati:
Cool, thanks for letting me know!

PR #709, opened 2019-03-07
Fixed outdated links

LorenzNickel:
Updated links


LorenzNickel:
I signed it!

eamonnmcmanus:
Thanks! Because of the way our source control works, we'll actually be applying this patch internally at Google, and it will appear in the next push.

PR #722, opened 2019-04-05
:memo: add how-to documentation for using auto-service with Gradle

ryandens:
I recently started using auto-service in a Gradle project. I noticed that while there was documentation on how to get started using auto-service with Maven, there was no documentation for Gradle. In my experience, having simple how-to guides for getting started lowers the barrier of entry for using this project. I wasn't sure if I should include a note that it appears that the `auto-service-annotations` artifact did not exist prior to version `1.0-rc5`. Any feedback is appreciated

### DIFF
--- a/service/README.md
+++ b/service/README.md
@@ -53,11 +53,30 @@ artifact as an "optional" dependency:
 <dependencies>
   <dependency>
     <groupId>com.google.auto.service</groupId>
+    <artifactId>auto-service-annotations</groupId>
+    <version>${version}</version>
+  </dependency>
+  <dependency>
+    <groupId>com.google.auto.service</groupId>
     <artifactId>auto-service</artifactId>
     <version>${version}</version>
     <optional>true</optional>
   </dependency>
 </dependencies>
+```
+Alternatively, instead of using the `optional` scope, you can add the dependency to the
+[annotationProcessorPaths](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths) 
+section of the configuration for the `maven-compiler-plugin`. 
+
+In a Gradle project, one would include the `auto-service-annotations`
+artifact as a `compileOnly` dependency and `auto-service`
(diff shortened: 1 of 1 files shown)


ryandens:
worth noting that auto-value has this same documentation: https://github.com/google/auto/blob/master/value/userguide/index.md

ryandens:
I signed it!

service/README.md
```diff
@@ -60,6 +60,17 @@ artifact as an "optional" dependency:
 </dependencies>
 ```
 
+In a Gradle project, one would include the `auto-service-annotations`
+artifact as a `compileOnly` dependency and `auto-service`
+as an `annotationProcessor`:
+
+```groovy
+dependencies {
+  compile             "com.google.auto.service:auto-service-annotations:${version}"
```
tbroyer: Should be `compileOnly` (per text above, and because the annotation has source retention)

ryandens: oops, meant to change that, totally agree!

tbroyer: Oh, fwiw, you could also use parentheses around the dependency. This is not really "idiomatic" in a Groovy DSL build, but it works and can also be copy/pasted in a Kotlin DSL file.

ryandens: works for me 👍I'll change the language of the code-block to kotlin, since that the language where the documentation is more idiomatic

ryandens:
I modified the Maven section to use the `auto-service-annotations` and a reference to the maven compiler plugin for using the annotationProcessorPaths. I didn't have a project in front of me where I could test out configuring the maven-compiler-plugin's annotationProcessorPaths, so I didn't give a copy-paste example in case it was incorrect. Feel free to add it to this branch if you want to

kostasdrakonakis:
It seems like that if you are in Kotlin project you should use `kapt` instead of `annotationProcessor` along with `apply plugin: 'kotlin-kapt`. Please do make sure you include that in the `README.md`

Issue #660, opened 2018-08-22
BasicAnnotationProcessor does not recursively validate superclasses / superinterfaces


Leland-Takamine:
`BasicAnnotationProcessor` only validates immediate superclass and superinterfaces. [This commit](https://github.com/Leland-Takamine/auto/commit/de950e8843210fe7dcd197891512d159cb77b2b8) demonstrates the behavior. The first test passes, but the second fails. It's possible to update `SuperficialValidation` to recursively validate superclasses and superinterfaces. In this something that `BasicAnnotationProcessor` should support?

ronshapiro:
I'm not positive, but I think here: https://github.com/google/auto/blob/9cc04ecb39166207a2835174b85ee209cc08aad0/common/src/main/java/com/google/auto/common/SuperficialValidation.java#L135

If we also check `validateElement(t.asElement())`, I think that should do the full type-hierarchy checking you're looking for.

Leland-Takamine:
So to clarify some options:

1. We can add `validateElement(t.asElement())` to `TYPE_VALIDATING_VISITOR.visitDeclared` to recursively validate superclasses, but this would introduce a cycle. To prevent an infinite loop, we can break the cycle by:
   a) Maintaining a set of visited elements. Sounds like it could work, but it would be a bit of refactoring.
   b) Removing `validateType` from `isValidBaseElement`. However, **this doesn't actually break the cycle** (ie: `ClassA self() { return this; }` would loop infinitely) so I don't think this would help.
2. Update `ELEMENT_VALIDATING_VISITOR.visitType` to call `validateElement` for each superclass and superinterface. This does not introduce a cycle like (1) since a class or interface cannot extend/implement itself.

ronshapiro:
@Leland-Takamine I tried option 2, but that is also stack-overflowing. Hrm.

Leland-Takamine:
@eamonnmcmanus @ronshapiro Pushed a PR up - let me know what you think.

TonyTangAndroid:
Is there any plan on how we are to proceed from here?  Been keeping on an eye on this issue for a while now.

PR #733, opened 2019-05-14
Recursive supertype and superinterface validation

Leland-Takamine:
Fixes google/auto#660

This change adds recursive validation for supertypes and superinterfaces to `SuperficialValidation`.


Leland-Takamine:
I signed it!

common/src/main/java/com/google/auto/common/SuperficialValidation.java
```diff
@@ -63,10 +64,13 @@ public static boolean validateElements(Iterable<? extends Element> elements) {
         }
 
         @Override public Boolean visitType(TypeElement e, Void p) {
+          TypeMirror superclass = e.getSuperclass();
           return isValidBaseElement(e)
               && validateElements(e.getTypeParameters())
               && validateTypes(e.getInterfaces())
-              && validateType(e.getSuperclass());
+              && validateType(superclass)
+              && validateElements(e.getInterfaces().stream().map(MoreTypes::asElement).collect(Collectors.toList()))
```
ronshapiro: Does `e.getInterfaces().stream(MoreTypes::asElement).allMatch(this::validateElement)`
 work?

common/src/test/java/com/google/auto/common/SuperficialValidationTest.java
```diff
@@ -199,6 +199,42 @@ public void handlesRecursiveType() {
         .compilesWithoutError();
   }
 
+  @Test
+  public void handlesRecursiveSuperinterface() {
+    JavaFileObject javaFileObject = JavaFileObjects.forSourceLines(
+        "test.TestClass",
+        "package test;",
+        "",
+        "interface TestClass implements TestClass {}");
```
ronshapiro: Was this failing before this change? This seems like an outright error that should never call annotation processing

common/src/test/java/com/google/auto/common/SuperficialValidationTest.java
```diff
@@ -276,6 +312,90 @@ void runAssertions() {
...
+        .failsToCompile();
+  }
+
+  @Test
+  public void missingGrandparentSuperclass() {
+    JavaFileObject javaFileObject = JavaFileObjects.forSourceLines(
+        "test.TestClass",
+        "package test;",
+        "",
+        "class Parent extends Missing {}",
```
ronshapiro: Can you use multiple classes here? That stays with Google's Java style

ronshapiro:
I'll try to run this through google-wide presubmit to check if anything breaks.

common/src/main/java/com/google/auto/common/SuperficialValidation.java
Leland-Takamine: I'll try that.

common/src/test/java/com/google/auto/common/SuperficialValidationTest.java
Leland-Takamine: This didn't fail before the change - just added coverage for this case since @eamonnmcmanus suggested doing so in #660. The test does validate that we don't reach annotation processing.

common/src/test/java/com/google/auto/common/SuperficialValidationTest.java
Leland-Takamine: Yep will update.

common/src/main/java/com/google/auto/common/SuperficialValidation.java
```diff
@@ -63,10 +64,13 @@ public static boolean validateElements(Iterable<? extends Element> elements) {
         }
 
         @Override public Boolean visitType(TypeElement e, Void p) {
+          TypeMirror superclass = e.getSuperclass();
           return isValidBaseElement(e)
               && validateElements(e.getTypeParameters())
               && validateTypes(e.getInterfaces())
-              && validateType(e.getSuperclass());
+              && validateType(superclass)
+              && e.getInterfaces().stream().map(MoreTypes::asElement).allMatch(SuperficialValidation::validateElement)
```
ronshapiro: If you want, `e -> validateElement(e)` is a little shorter than the method reference

Leland-Takamine: Would prefer to leave as is - `e` conflicts with the existing method parameter and the IDE produces a warning with the lambda.

ronshapiro:
Unfortunately this is introducing some cycles. Here's a repro:

```java
@Component
class Foo extends Reference {}
```

Looks like there's a cycle between `java.lang.Reference` and `java.lang.ReferenceQueue`.

I'm going to think more, but I think we may want to use something like `com.google.common.graph.Traverser` instead so that we can ensure we never infinitely recur.

Leland-Takamine:
@ronshapiro Would it be sufficient to just keep track of visited `Elements` and skip those we've already seen?

ronshapiro:
We could try that. I think passing those around may be icky, so that's why I suggested `Traverser`, which can handle that for us. It abstracts "given this isntance, these are the other things that need to be checked" from "check this individual thing".

Leland-Takamine:
Wouldn't we need to handle cycles manually when building to graph in order to leverage `Traverser`?

Leland-Takamine:
> Wouldn't we need to handle cycles manually when building to graph in order to leverage `Traverser`?

Ah never mind - I see now that [Traverser.forGraph](https://google.github.io/guava/releases/27.0-jre/api/docs/src-html/com/google/common/graph/Traverser.html#line.92) accepts a function not a graph.

Leland-Takamine:
Updated to handle cycles by keeping track of visited Elements at the instance level. This requires making all internal logic non-static. Public static API is unchanged.

Leland-Takamine:
@ronshapiro Mind taking another look?

ronshapiro:
I'll send this through our tests again and report back

ronshapiro:
I'm seeing a few cases of this come up now:

`java.lang.IllegalArgumentException: <nulltype> cannot be represented as a Class<?>.`

From within

`SuperficialValidation$3.defaultAction`

Leland-Takamine:
@ronshapiro Are you able to track down which tests are failing? I'd like to figure out a repro case so I can write a test for that.

ronshapiro:
I haven't had time yet to investigate

Leland-Takamine:
@ronshapiro I've taken a look at the code but I'm still unclear on how we'd be visiting a null type. It would be really helpful to see the failing tests to help debug.

Leland-Takamine:
Ping on this :)

PR #764, opened 2019-10-01
Update to Truth 1.0

elharo:
Going to need a little more work.


cpovirk:
Typically updating to Truth 1.0 just requires updating Compile-Testing to 0.18 in projects that use that. Hopefully that's all it takes here.

PR #765, opened 2019-10-02
Mark autovalue-annotations optional (Maven) or compileOnly (gradle)

elharo:
@eamonnmcmanus @saturnism

### DIFF
--- a/value/userguide/index.md
+++ b/value/userguide/index.md
@@ -83,6 +83,7 @@ Maven users should add the following to the project's `pom.xml` file:
   <groupId>com.google.auto.value</groupId>
   <artifactId>auto-value-annotations</artifactId>
   <version>1.6.6</version>
+  <optional>true</optional>
 </dependency>
 <dependency>
   <groupId>com.google.auto.value</groupId>
@@ -101,8 +102,8 @@ Gradle users can declare the dependencies in their `build.gradle` script:
 
 ```groovy
 dependencies {
-  // Use 'api' rather than 'compile' for Android or java-library projects.
-  compile             "com.google.auto.value:auto-value-annotations:1.6.6"
+  // Use 'api' rather than 'compileOnly' for Android or java-library projects.
+  compileOnly         "com.google.auto.value:auto-value-annotations:1.6.6"
   annotationProcessor "com.google.auto.value:auto-value:1.6.6"
 }
 ```


value/userguide/index.md
```diff
@@ -83,6 +83,7 @@ Maven users should add the following to the project's `pom.xml` file:
   <groupId>com.google.auto.value</groupId>
   <artifactId>auto-value-annotations</artifactId>
   <version>1.6.6</version>
+  <optional>true</optional>
```
eamonnmcmanus: I'm not sure this is a completely safe recommendation. Since `@AutoValue` has `CLASS` retention, when you compile `@AutoValue class Foo` there is a reference to `com.google.auto.value.AutoValue` in `Foo.class`, even if that reference is not visible to reflection. If a tool, such as another annotation processor, starts looking at the annotations referenced in `Foo.class` it may be perturbed by the missing reference.

Meanwhile, `auto-value-annotations` is a pretty tiny dependency with no transitive dependencies. So I'm not sure there's much incentive to making people have to think about whether to mark it optional or not.

elharo: auto-value-annotations does seem to be causing us trouble in javacore and possibly pulling in autoservice, so retaining this in the classpath isn't guaranteede4ly safe. Why does this have class retention? This would seem to contradict the claim that "Your choice to use AutoValue is essentially API-invisible."

eamonnmcmanus: "Possibly pulling in autoservice", that's surprising to me. I don't see where that reference would come from. AutoService is referenced by the AutoValue processor but not by the annotations in this jar.

We did have `SOURCE` retention initially, but it turns out that some tools want to see whether a class is an `@AutoValue`. For example, AutoValue itself does not allow one `@AutoValue` class to extend another, but with `SOURCE` retention it would only be able to check that if the two classes were being compiled at the same time.

tbroyer: > I'm not sure this is a completely safe recommendation.

:100: 
This is similar to ErrorProne annotations, Checker Framework annotations, etc. that had been causing issues when absent from the classpath (generally when the downstream project is using ErrorProne IIRC), and led to Guava (among others) to declare them as "normal" dependencies.

> We did have `SOURCE` retention initially, but it turns out that some tools want to see whether a class is an `@AutoValue`.

FWIW, ErrorProne is one of those tools: https://errorprone.info/bugpattern/ExtendsAutoValue

TL;DR: `@AutoValue` is somehow part of your API, and should be included in the classpath **by default**. Feel free to exclude it from your transitive dependencies if you don't want it though.

elharo: Yes, I haven't figured that out either. It's a weird one and I may be guessing wrong about the root cause. However I did find that marking autovalue-annotations optional made the AutoService error go away.

cpovirk: A couple threads that discuss why we moved away from `optional`/`provided` for Guava's annotation dependencies: https://github.com/google/guava/issues/2824, https://github.com/google/guava/issues/2721

saturnism: one of the last comment in google/guava#2721 was to prefer as optional, also pointed to https://globalmentor.atlassian.net/browse/JAVA-47

If another user of autovalue tries to extend on an autovalue class, that user would've probably included autovalue related libraries in the compile classpath in the first place, right? In which case, the depended lib still doesn't need to export it. I feel Runtime retention is fine, but the absence of it in classpath should only cause issues if when it needs to be used/reflected on.

tbroyer: > If another user of autovalue tries to extend on an autovalue class, that user would've probably included autovalue related libraries in the compile classpath in the first place, right? In which case, the depended lib still doesn't need to export it.

Wrong. See discussion in https://github.com/google/error-prone/pull/1354#discussion_r327483560

> I feel Runtime retention is fine, but the absence of it in classpath should only cause issues if when it needs to be used/reflected on.

I think we all agree if we're talking about the runtime classpath, but the issue here is the compile classpath. Unfortunately we cannot say "include this in the compile classpath (of downstream projects consuming AutoValue) but not in the runtime classpath". So either you live with a small unnecessary jar in your runtime classpath, or you explicitly exclude it.

saturnism: my bad - so the check is coming from error prone. if error prone wants to check autovalue, shouldn't error prone be the party responsible for brining in the autovalue annotation dependency? and since error_prone_core runs within a plugin, its dependencies would be limited to the plugin classloader.

if there is just one version of this, i really don't mind it... but when this is included everywhere by foundational libraries, i really feel it gets excessive very quickly - it already has.

cpovirk: My understanding is that plugins like Error Prone have their own classpath, and types from that classpath aren't available to the check logic.

Now, as it turns out, the specific Error Prone check `ExtendsAutoValue` appears not to need `@AutoValue` to be present on the "normal" classpath: It can operate simply based on the name of the type, without knowing anything more about it ([1](https://github.com/google/error-prone/blob/32a5730052d4173195bb45fec8fce10a9d8b8822/core/src/main/java/com/google/errorprone/bugpatterns/ExtendsAutoValue.java#L48), [2](https://github.com/google/error-prone/blob/32a5730052d4173195bb45fec8fce10a9d8b8822/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L679), [3](https://github.com/google/error-prone/blob/32a5730052d4173195bb45fec8fce10a9d8b8822/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L722)).

This is not necessarily the case in general: First, some processors might make the (usually reasonable) assumption that a type used by the code being compiled _is_ available. Second, some annotations are tagged as "inherited," and checks need to be able to read the class file to determine whether they are ([1](https://github.com/google/error-prone/blob/32a5730052d4173195bb45fec8fce10a9d8b8822/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L682), [2](https://github.com/google/error-prone/blob/32a5730052d4173195bb45fec8fce10a9d8b8822/check_api/src/main/java/com/google/errorprone/util/ASTHelpers.java#L701)).

(It's also still possible that it would make sense for `auto-value-annotations` to stop releasing a new version automatically every that `auto-value` does. Oh, and how much does the Linkage Checker help us to distinguish between usually harmless conflicts (like we're seeing with `auto-value`) and harmful ones (where the classes actually change between versions)? Or is there actual harm in this case, in which case the change wouldn't help?)

saturnism: > My understanding is that plugins like Error Prone have their own classpath, and types from that classpath aren't available to the check logic.

I feel this is worth a try, and most likely works, since plugin runtime classpath contains errorprone plugin, and can contain more dependencies specific to the runtime of the plugin.

> This is not necessarily the case in general: First, some processors might make the (usually reasonable) assumption that a type used by the code being compiled is available

if the processor needs to use an annotation, user should place those annotations on their own classpath either as optionally scoped, or in the plugin runtime scoped, or as a direct dependency of the plugin...

> It's also still possible that it would make sense for auto-value-annotations to stop releasing a new version automatically every that auto-value does

+1 for this one :) i'd also prefer this.

cpovirk: To be explicit: I'm pretty sure that putting the annotations as a dependency of the *plugin* does *not* let the plugin read them through the plugin APIs. The annotations need to be on the compile classpath. So that leaves us with this:

> user should place those annotations on their own classpath either as optionally scoped

That way should work.

(I'm still a bigger fan of avoiding `optional` here, but the main thing I want is to make sure we're clear about what users need to do to get the behavior they want.)

saturnism: i did give it a try; and it certainly didn't work for error-prone-annotations.

for autovalue, i haven't been able to trigger `ExtendsAutoValue` violation w/ or w/o the `auto-value-annotations`. will test more out.

cpovirk: `ExtendsAutoValue` is a warning by default. I'm not sure how that's surfaced in open-source builds. Internally, you'd probably need to open a CL in Critique after the analyzer has run.

Alternatively, have a look at my CL 273372038, in which I promoted it to an error. You can build `java/com/google/bar` to see a failure (even though I've hidden the internal equivalent of `auto-value-annotations` from that compilation).

cpovirk: With these docs changed in https://github.com/google/auto/commit/43ff5f2d34e7ac88bd0722281d07303618c85a07, maybe we can now close this PR?

Or maybe it can be revised to still update the Gradle section? Nowadays, we can even recommend [`compileOnlyApi`](https://docs.gradle.org/6.7-rc-1/release-notes.html#compile-only-api).

elharo:
https://github.com/google/auto/issues/995

PR #789, opened 2019-11-14
Exclude unused transitive dependencies

cesarsotovalero:
Hello, I noticed that dependencies `com.google.guava:listenablefuture` and `com.google.auto.value:auto-value-annotations` are included in the dependency tree of module `auto-common`. However, these transitive dependencies are not used and therefore can be excluded in the `pom`. This makes the library slimmer for its users and the Maven dependency clearer and less complex.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -69,6 +69,12 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>${guava.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.guava</groupId>
+          <artifactId>listenablefuture</artifactId>          
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <!-- Used only by GeneratedAnnotationSpecs.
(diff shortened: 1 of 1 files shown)


tbroyer:
Do you really want to start maintaining exclusion lists for things that cause absolutely zero problem? `listenablefuture` is a single class file (iirc), and `auto-value-annotations` are only in the `test` scope!

JakeWharton:
ListenableFuture is actually empty for the JRE Guava dep, but it's presence in the dependency tree is vital as it overrides the version which does have a class.

cpovirk:
As JakeWharton says, the dependency on `listenablefuture-9999` is [a hacky way](https://groups.google.com/d/topic/guava-announce/Km82fZG68Sw/discussion) to prevent JPMS module conflicts. Excluding the dependency could reintroduce those conflicts. (And yes, `listenablefuture-9999` itself is an empty jar, since the single class file in `listenablefuture-1.0` (which tbroyer is referring to) is part of Guava itself.)

The dependency on `auto-value-annotations`, as tbroyer notes, is a test-only dependency. Excluding it might be safe in this case, but [sometimes excluding annotation-only dependencies can cause problems](https://github.com/google/guava/wiki/Release22#additional-changes), so even in the case of prod dependencies, we generally prefer to keep them and let end users exclude them when desired.

PR #814, opened 2020-02-25
Update AutovalueProcessor to be "ISOLATING" as the other processors

JunYang-CD:
Issue detected during building with KAPT for kotlon android project.
[WARN] Incremental annotation processing requested, but support is disabled because the following processors are not incremental: 
com.google.auto.value.processor.AutoValueProcessor (DYNAMIC)

Thanks for this commit from @tbroyer
https://github.com/google/auto/commit/a5673d06f687e1354f1f069cce36136538cf532c#diff-3c05eef48fd8e04c525c96b18fab9281R66

But not sure why the AutoValueProcessor was annoated with @IncrementalAnnotationProcessor(IncrementalAnnotationProcessorType.DYNAMIC)
instead of @IncrementalAnnotationProcessor(IncrementalAnnotationProcessorType.ISOLATING)as the others.

Having tested it by setting it as "ISOLATING", the kapt doesn't compain it blocks incremental build anymore.

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
+++ b/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java
@@ -63,7 +63,7 @@ import net.ltgt.gradle.incap.IncrementalAnnotationProcessorType;
  */
 @AutoService(Processor.class)
 @SupportedAnnotationTypes(AUTO_VALUE_NAME)
-@IncrementalAnnotationProcessor(IncrementalAnnotationProcessorType.DYNAMIC)
+@IncrementalAnnotationProcessor(IncrementalAnnotationProcessorType.ISOLATING)
 public class AutoValueProcessor extends AutoValueOrOneOfProcessor {
   private static final String OMIT_IDENTIFIERS_OPTION = "com.google.auto.value.OmitIdentifiers";
 


tbroyer:
As you can see in the linked commit, the actual incremental type is determined dynamically from the AutoValueExtensions in use.
Could it be that Kapt doesn't support the whole spec for incremental processing and doesn't support dynamic processors? It shouldn't report "DYNAMIC" but the actual, computed incremental type.

tbroyer:
Either that or maybe you have an AutoValueExtension that has not been updated with incremental support and triggers the `UNKNOWN` code path.

JunYang-CD:
> As you can see in the linked commit, the actual incremental type is determined dynamically from the AutoValueExtensions in use.
> Could it be that Kapt doesn't support the whole spec for incremental processing and doesn't support dynamic processors? It shouldn't report "DYNAMIC" but the actual, computed incremental type.

Kapt do supports "aggregating", "isolating" and "dynamic", there are couple  our internal annotation processors declared as "dynamic" too.

JunYang-CD:
> Either that or maybe you have an AutoValueExtension that has not been updated with incremental support and triggers the `UNKNOWN` code path.

It is AutoValue 1.7 used in my project. Having checked the code, It looks like that MemoizeExtension which derives from AutoValueExtension is the only extension applied to the AutoValueProcessor. Can you please tell me how to list out all the extensions that is applied to the AutoValueProcessor if it's not the case? I can't any possible cause resulting in  incremental type "UNKOWN"

JunYang-CD:
@tbroyer I will close this PR soon. You are correct that there is another auto extension in our project which doesn't support incremental build. It is using the old version 0.2.6 of (rharter/auto-value-parcel). Thanks for your active response.

Issue #813, opened 2020-02-24
@AutoService to generate META-INF/gradle-plugins as well for standalone gradle plugins


kostasdrakonakis:
When creating standalone `Gradle` plugin we have to generate some files under `META-INF/gradle-plugins` .

I understand that the problem arises when you have to pick the correct name for the custom plugin.

I was thinking that maybe we can add an argument `pluginName`  to the `@AutoService` annotation or we can add another annotation like `@Plugin` where we could specify the name.

Example:

* pluginName

```kotlin
package com.example

@AutoService(value= [Project::class], pluginName = "com.domain.my_custom_plugin")
class CustomStandAlonePlugin : Plugin<Project> {
```
* different annotation

```kotlin
package com.example

@AutoService(Project::class)
@Plugin("com.domain.my_custom_plugin")
class CustomStandAlonePlugin : Plugin<Project> {
```

So when the processor will generate it should generate a file under `META-INF/gradle-plugins`:

`com.domain.my_custom_plugin.properties`

and the entry to this file should be:

`implementation-class=com.example.CustomStandAlonePlugin`

Is there any intention on adding the above feature to auto service?

vlsi:
Gradle has Precompiled Script Plugins: https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins

In other words, if you name a file like `src/main/kotlin/java-library-convention.gradle.kts`, then Gradle automatically generates metadata to make it a `java-library-convention` plugin (it generates `META-INF/gradle-plugins/java-library-convention.properties`)

tbroyer:
The `java-gradle-plugin` will also generate the plugin descriptors from DSL in the build script: https://docs.gradle.org/current/userguide/java_gradle_plugin.html

vlsi:
In other words, I suggest closing the issue as "not planned" as Gradle provides enough services to generate `META-INF/gradle-plugins`.

PR #822, opened 2020-03-11
Add support for generating META-INF for standalone gradle plugins

kostasdrakonakis:


kostasdrakonakis:
Resolves #813

kostasdrakonakis:
guys any plans to review this? I am really looking forward for this feature as it will save us some time when we create multiple gradle plugins

service/README.md
```diff
@@ -42,6 +42,35 @@ automatically load it, and include it in its normal annotation processing
...
+
+@AutoService(Plugin.class)
+@PluginName("com.example")
+final class MyPlugin implements Plugin<Project> {
+  // …
+}
+```
+
+AutoService will generate the file
+`META-INF/gradle-plugins/com.example.properties` in the output classes
```
AnEmortalKid, line 66: Since AutoService sort of lands things in `META-INF/services`, would it be better to have an `@AutoGradlePlugin` annotation / corresponding processor ?

kostasdrakonakis: As a matter of fact it actually makes sense.

I am not very good with names and `PluginName` does not really express the name but rather the id. My original thought was to name the annotation `PluginId` but because Gradle already has classes named as `Plugin` and `PluginId` naming the annotation as such would make it confusing for having to import the fully qualified name every time.

ZacSweers:
@kostasdrakonakis the gradle's `java-gradle-plugin` plugin handles this for you already. I think this makes more sense as a standalone project

AnEmortalKid:
>  I think this makes more sense as a standalone project

+1 , should be fairly easy to make a standalone project

tbroyer:
I'd go as far as saying it doesn't make any sense to add such feature to auto-service.

Auto-service:
* generates `META-INF/services/xxx`
* ensures the annotated class implements/extends `xxx`
* aggregates all annotated classes with the same annotation argument (the `xxx`) into the same output file

The requirements for the proposed feature:
* generates a `META-INF/gradle-plugins/xxx.properties`
* does **not** generate `META-INF/services/org.gradle.api.Plugin` (it probably could, but shouldn't, as Gradle plugins don't follow the ServiceLoader requirements: they can have @Inject`-annotated constructors for instance)
* ensures uniqueness of the `xxx` among annotated classes

kostasdrakonakis:
> @kostasdrakonakis the gradle's `java-gradle-plugin` plugin handles this for you already. I think this makes more sense as a standalone project

ok I agree.

> I'd go as far as saying it doesn't make any sense to add such feature to auto-service.
> 
> Auto-service:
> 
> * generates `META-INF/services/xxx`
> * ensures the annotated class implements/extends `xxx`
> * aggregates all annotated classes with the same annotation argument (the `xxx`) into the same output file
> 
> The requirements for the proposed feature:
> 
> * generates a `META-INF/gradle-plugins/xxx.properties`
> * does **not** generate `META-INF/services/org.gradle.api.Plugin` (it probably could, but shouldn't, as Gradle plugins don't follow the ServiceLoader requirements: they can have @Inject`-annotated constructors for instance)
> * ensures uniqueness of the `xxx` among annotated classes

Regarding your points as of it does not make sense, I would say it does make sense as there are multiple times that we have an annotation processor along with a gradle plugin often at same project(different modules though) so it makes sense if one uses auto-service to have both features bundled.

A separate project would have to do extra job in order to ensure that the annotated class implements/extends the `Plugin` interface while the `@AutoService` already handles that.

Another reason would be since that both annotations `@AutoService` and `@AutoGradlePlugin` generate files under META-INF I thought it would be nice to have them in 1 project.

At the end I agree about standalone project so how do you suggest this should go?

Create another directory under `auto/`(e.g. `auto/gradle-plugin` just like `auto/service`) and move the implementation with its own processor there or to separate repository?

ZacSweers:
It should be its own project in its own repository

kostasdrakonakis:
I moved it to a separate repository here is the link if you want to check it: https://github.com/Netscal/autogradleplugin

So now the PR can be closed I guess

PR #837, opened 2020-04-21
Ensure to process all providers in when generating config files

bbyk:
In `AutoServiceProcessor#generateConfigFiles` the loop over providers can
be left prematurely if for a provider type there is no new services. The
remaining provider types can still have new services though.


netdpb:
Closing this PR since #839 describes the problem.

Issue #427, opened 2017-01-17
Memoized extension not picking up error-prone's `@LazyInit` annotations on the classpath


ZacSweers:
Since `@Memoized` results in non-final fields, it can't be used in `@Immutable`-annotated classes. error-prone does have support for `@LazyInit` on fields like this though, would it be possible to have some sort of opt-in mechanism and generate these annotations with the fields?

ronshapiro:
It should be doing that: https://github.com/google/auto/blob/57dfad360306619a820e6aae4a14a1aa67c29299/value/src/main/java/com/google/auto/value/extension/memoized/MemoizeExtension.java#L352

Are you not seeing it happen that way?

tbroyer:
@ronshapiro AFAICT it hasn't been released. The commit is from November and the last release on Central is from October: https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.google.auto.value%22%20AND%20a%3A%22auto-value%22

[GitHub releases page](https://github.com/google/auto/releases) says there's been a 1.4-rc-2 release (and there's a tag for it) that should contain the commit but it's apparently not on Central.

ronshapiro:
That's weird, I do see it here: https://oss.sonatype.org/content/repositories/public/com/google/auto/value/auto-value/1.4-rc2/

tbroyer:
It has apparently not been sync'd from OSSRH to Central: https://repo1.maven.org/maven2/com/google/auto/value/auto-value/  
Contact Sonatype?

hzsweers:
Yeah I'm using -rc2 as well and it still wasn't showing up (haven't had maven issues). I'll tinker a bit more later today

hzsweers:
I still can't get this to work for some reason

eamonnmcmanus:
I assume this is fixed with AutoValue 1.4. Please let me know if not and I will reopen.

hzsweers:
Yes I still see this issue on 1.4 :/. The LazyInit annotations never get added. I'll try to put together a repro case

hzsweers:
Repro case: https://github.com/hzsweers/json-serialization-benchmarking/tree/z/errorProneAnnotations

Steps:
- Clone and check out the `z/errorProneAnnotations` branch
- `./gradlew compileJava`
- The class using `@Memoized` is `FriendAV.java`, for its `toString()` and `hashCode()` impls.
[...]

PR #861, opened 2020-07-08
MOE Sync 2020-07-08

kevinb9n:
This code has been reviewed and submitted internally. Feel free to discuss on
the PR, and we can submit follow-up changes as necessary.

Commits:
=====
<p> Stop the LazyInit annotation from getting shaded by Maven, so that AutoValue can find it on the classpath.

The @LazyInit annotation is added if it's found on the classpath (https://github.com/google/auto/blob/da84ef1fae38f2d72901c2b95674271e89600f28/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java#L577) but shading rewrites the package at https://github.com/google/auto/blob/da84ef1fae38f2d72901c2b95674271e89600f28/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java#L104 from "com.google.errorprone.annotations.concurrent" to "autovalue.shaded.com.google$.errorprone.annotations.$concurrent" (you can verify this by disassembling the bytecode for MemoizeExtension.class in the release JAR). This means that even with the Error Prone annotations on the classpath, it won't be able to find the LazyInit annotation.

Splitting up the package with a call to String#concat means Maven will no longer rewrite it, so AutoValue will be able to find the LazyInit annotation on the classpath if it's there.

Fixes https://github.com/google/auto/issues/427

RELNOTES=Stop the LazyInit annotation from getting shaded by Maven, so that AutoValue can find it on the classpath.

74a18fd500075429c07b95daec79fa88e27a1296

### DIFF
--- a/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java
+++ b/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java
@@ -100,8 +100,11 @@ public final class MemoizeExtension extends AutoValueExtension {
   private static final String AUTO_VALUE_NAME = AUTO_VALUE_PACKAGE_NAME + "AutoValue";
   private static final String COPY_ANNOTATIONS_NAME = AUTO_VALUE_NAME + ".CopyAnnotations";
 
+  // Maven is configured to shade (rewrite) com.google packages to prevent dependency conflicts.
+  // Split up the package here with a call to concat to prevent Maven from finding and rewriting it,
+  // so that this will be able to find the LazyInit annotation if it's on the classpath.
   private static final ClassName LAZY_INIT =
-      ClassName.get("com.google.errorprone.annotations.concurrent", "LazyInit");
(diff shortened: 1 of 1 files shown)


RichardFourie:
This "Fix" now introduced the lovely side effect of:
@AutoValue processor threw an exception: org.eclipse.jdt.internal.compiler.problem.AbortCompilation: Pb(324) The type com.google.errorprone.annotations.concurrent.LazyInit cannot be resolved. It is indirectly referenced from required .class files

eamonnmcmanus:
The change was made last July but nobody has complained about it until now. Furthermore, the AutoValue `.class` files don't in fact reference the `LazyInit` type, either before or after the change. AutoValue will attempt to see if the class is available [here](https://github.com/google/auto/blob/89c0e53ca75ccf894d251a2223e5a18e4d6db2f7/value/src/main/java/com/google/auto/value/extension/memoized/processor/MemoizeExtension.java#L584), but that's a reference from code, not from `.class` files. Are you sure the problem is with AutoValue?

PR #864, opened 2020-07-27
Fully Qualify @Override to avoid name conflicts

kpayson64:
This works around the issue where Override is defined in a user's namespace.


eamonnmcmanus:
This is great, thanks! But since you're a Googler, it would actually be more straightforward if you sent it as an internal code review. We sync from Google's source repo to GitHub rather than the other way around.

Issue #881, opened 2020-09-21
Null-Annotations for factory


frankbenoit:
Hi,

is there a way to generate the factories with null-annotations, like the Eclipse JDT @NonNull or @NonNullByDefault?

Franks

frankbenoit:
An implementation might work like this:

AutoFactoryProcessor.getSupportedOptions configures to allow the option "nullAnnotations".
For now the value "JDT" is accepted.

If the javac option `-AnullAnnotations=JDT` is given, then the `@org.eclipse.jdt.annotations.NonNullByDefault` is attached to the factory,

```java
@org.eclipse.jdt.annotation.NonNullByDefault
public class MyClassFactory ...
```

eamonnmcmanus:
Maybe we could do something like [`@AutoValue.CopyAnnotations`](https://www.javadoc.io/doc/com.google.auto.value/auto-value/latest/com/google/auto/value/AutoValue.CopyAnnotations.html). If your `@AutoFactory` class has `@AutoFactory.CopyAnnotations` then any other annotations on it get copied to the implementation. So the source class

```
@AutoFactory
@AutoFactory.CopyAnnotations
@NonNullByDefault
class Whatever {...}
```

would result in the generated class

```
@NonNullByDefault
class WhateverFactory {...}
```

I think that might be better than hardwiring logic about specific annotations.

frankbenoit:
I agree that a more general mechanism would be good.

However, your example for @NonNullByDefault makes the assumption, that the user-class and the factory are always both @NonNullByDefault. But this is not necessarily true. As a user I want to be free, to not have that annotation on my class, but the factory shall have it.

frankbenoit:
I added a proposal implementation for the textual way, see PR #884 

Current limitation:
- no annotation arguments
- FactoryAnnotation is not nested in AutoFactory

ronshapiro:
Related internal bug: b/31342163

eamonnmcmanus:
I'm not sure there's a really good solution here. I did think about a solution similar to #884 but, as @frankbenoit mentions, that doesn't allow annotation arguments, and also requires people to spell out the fully-qualified annotation name in a string constant (since you can't write `@FactoryAnnotation(MyAnnotation.class.getName())`). I guess the second problem could be addressed by making the annotation look like this?

```
public @interface FactoryAnnotation {
  Class<? extends Annotation> value();
}
```

And maybe we could allow for parameters in a slightly clunky way like this:

```
[...]

PR #885, opened 2020-10-02
Add custom factory annotation via class

frankbenoit:
Signed-off-by: Frank Benoit <frank.rene.benoit@gmail.com>


frankbenoit:
See #881

frankbenoit:
decline in favor of follow up proposal

PR #888, opened 2020-10-09
Add CopyAnnotations

frankbenoit:
proposal for #881 with @CopyAnnotations

Signed-off-by: Frank Benoit <frank.rene.benoit@gmail.com>

### DIFF
--- a/factory/src/main/java/com/google/auto/factory/processor/AutoFactoryDeclaration.java
+++ b/factory/src/main/java/com/google/auto/factory/processor/AutoFactoryDeclaration.java
@@ -26,6 +26,7 @@ import static javax.lang.model.util.ElementFilter.typesIn;
 import static javax.tools.Diagnostic.Kind.ERROR;
 
 import com.google.auto.factory.AutoFactory;
+import com.google.auto.factory.AutoFactory.CopyAnnotations;
 import com.google.auto.value.AutoValue;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
@@ -36,6 +37,7 @@ import com.google.common.collect.ImmutableSet;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import javax.annotation.processing.Messager;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.AnnotationMirror;
@@ -43,6 +45,8 @@ import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.MirroredTypeException;
+import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
 import javax.lang.model.util.Elements;
 
@@ -61,6 +65,7 @@ abstract class AutoFactoryDeclaration {
   abstract boolean allowSubclasses();
   abstract AnnotationMirror mirror();
   abstract ImmutableMap<String, AnnotationValue> valuesMap();
+  abstract List<CopyAnnotationsDescriptor> factoryAnnotations();
 
   PackageAndClass getFactoryName() {
     String packageName = getPackage(targetType()).getQualifiedName().toString();
@@ -88,8 +93,10 @@ abstract class AutoFactoryDeclaration {
   static final class Factory {
     private final Elements elements;
     private final Messager messager;
+    private Elements elementUtils;
 
-    Factory(Elements elements, Messager messager) {
+    Factory(Elements elementUtils, Elements elements, Messager messager) {
+      this.elementUtils = elementUtils;
       this.elements = elements;
       this.messager = messager;
     }
@@ -118,7 +125,7 @@ abstract class AutoFactoryDeclaration {
       if (extendingType == null) {
         messager.printMessage(ERROR, "Unable to find the type: "
             + extendingValue.getValue().toString(),
-                element, mirror, extendingValue);
+            element, mirror, extendingValue);
         return Optional.absent();
       } else if (!isValidSupertypeForClass(extendingType)) {
         messager.printMessage(ERROR,
@@ -135,7 +142,7 @@ abstract class AutoFactoryDeclaration {
                   return constructor.getParameters().isEmpty();
                 }
               })
-              .toList();
+          .toList();
       if (noParameterConstructors.size() == 0) {
         messager.printMessage(ERROR,
             String.format("%s is not a valid supertype for a factory. "
@@ -157,6 +164,8 @@ abstract class AutoFactoryDeclaration {
       AnnotationValue allowSubclassesValue = checkNotNull(values.get("allowSubclasses"));
       boolean allowSubclasses = AnnotationValues.asBoolean(allowSubclassesValue);
 
+      List<CopyAnnotationsDescriptor> factoryAnnotations = getFactoryAnnotations(element);
+
       return Optional.<AutoFactoryDeclaration>of(
           new AutoValue_AutoFactoryDeclaration(
               getAnnotatedType(element),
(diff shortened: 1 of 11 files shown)


eamonnmcmanus (changes requested):
I think this looks good overall. However, for a change like this I'm going to want to go through Google's internal API review process before committing. We won't be able to change the API easily afterwards, so we want to be confident we'll be happy with it. We can continue the discussion here, but it is at least _possible_ that the outcome of the API discussion will be a different suggested direction.

factory/src/test/resources/good/ClassWithCopyAnnotations.java
```diff
@@ -0,0 +1,33 @@
...
+package tests;
+
+import com.google.auto.factory.AutoFactory;
+import com.google.auto.factory.CopyAnnotations;
+import com.google.errorprone.annotations.Immutable;
+import org.checkerframework.framework.qual.DefaultQualifier;
+import org.checkerframework.framework.qual.DefaultQualifiers;
+import javax.inject.Singleton;
+
+@CopyAnnotations
```
eamonnmcmanus, line 25: I think the annotation should apply only to the class that has the `@AutoFactory` annotation. Otherwise it's hard to know what the eventual result will be. What would happen if `ClassWithCopyAnnotations` had its own `@CopyAnnotations` annotation, for example?

factory/src/main/java/com/google/auto/factory/processor/AutoFactoryDeclaration.java
```diff
@@ -166,7 +175,54 @@ PackageAndClass getFactoryName() {
...
+          .filter(unequalTo(CopyAnnotations.class))
+          .collect(Collectors.toList());
+    }
+
+    private <T> Predicate<? super AnnotationMirror> unequalTo( Class<T> type ){
+      return am -> !type.getCanonicalName().equals(am.getAnnotationType().toString());
+    }
+    private List<? extends AnnotationMirror> getAllAnnotations(Element element, CopyAnnotations fa) {
+      TypeMirror type = extractValueOfClassType(fa);
+      if( !"java.lang.Object".equals(type.toString())) {
```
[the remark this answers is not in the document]
frankbenoit: I tried it with command-line tool and eclipse plugin, both re-format the whole file to have many lines changes, that I did not touch.
How exactly do you use it?

factory/src/test/resources/good/ClassWithCopyAnnotations.java
frankbenoit: In this example the @AutoFactory is placed at the ctor.
Do you say, you want that in this case @AutoFactory.CopyAnnotations is not allowed? Only if the @AutoFactory is placed really at the class?

factory/src/main/java/com/google/auto/factory/processor/AutoFactoryDeclaration.java
```diff
@@ -166,7 +175,54 @@ PackageAndClass getFactoryName() {
...
+    }
+
+    private List<CopyAnnotationsDescriptor> getFactoryAnnotations(Element element) {
+      
+      if( element instanceof ExecutableElement ) {
+        return getFactoryAnnotations(element.getEnclosingElement());
+      }
+      ImmutableList.Builder<CopyAnnotationsDescriptor> resultBuilder = ImmutableList.builder();
+      
+      CopyAnnotations fa = element.getAnnotation(CopyAnnotations.class);
```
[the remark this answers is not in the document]
frankbenoit: The auto-factory does have this direct reference to the annotation in 5 locations.
Perhaps this can be done in a different change in future, and then do it for all of those references?

factory/src/main/java/com/google/auto/factory/CopyAnnotations.java
```diff
@@ -0,0 +1,66 @@
...
+ * public class Foo { ...
+ *     &#064;NonNullByDefault
+ *     public static class Source {}
+ * }
+ * </pre>
+ * 
+ * @author Frank Benoit
+ */
+@Target({ TYPE })
+public @interface CopyAnnotations {
```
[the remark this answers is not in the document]
frankbenoit: done

factory/src/test/java/com/google/auto/factory/processor/AutoFactoryProcessorTest.java
```diff
@@ -434,6 +434,28 @@ public void defaultPackage() {
         .generatesSources(loadExpectedFile("expected/DefaultPackageFactory.java"));
   }
 
+  @Test
+  public void classWithCopyAnnotations() {
+    JavaFileObject file = JavaFileObjects.forResource("good/ClassWithCopyAnnotations.java");
```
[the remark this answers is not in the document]
frankbenoit: Yes, I know the pain of reference-compare tests and when it goes to 100s of files. Not yet a problem here :-)

Why not divide and conquer? Split testing at the `FactoryDescriptor`.
1. Suppose the `FactoryWriter` is already seen as tested reliably (or is tested separately).
Then other tests might get access to the `FactoryDescriptor` that is passed into the FactoryWriter. 
This `FactoryDescriptor` contains all information needed for tests, to make them less over specified (as a text compare is).
Then it would be easy to add specific tiny unit tests, testing exactly one thing/behavior.
2. Write the java source text in unit test directly (hopefully soon with text blocks support). So the test source and asserts are together. Perhaps evolve an api to compose the java source texts with less duplication.

factory/src/test/resources/good/ClassWithCopyAnnotations.java
frankbenoit: @eamonnmcmanus can you please comment on this

factory/src/main/java/com/google/auto/factory/processor/AutoFactoryDeclaration.java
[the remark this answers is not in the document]
eamonnmcmanus: OK

factory/src/test/resources/good/ClassWithCopyAnnotations.java
eamonnmcmanus: My earlier comment wasn't quite right. I should have said that `@AutoFactory.CopyAnnotations` should be applied to exactly the same element as `@AutoFactory`. If `@AutoFactory` is applied to a class then `@CopyAnnotations` should be applied to the same class. If `@AutoFactory` is applied to a constructor then `@CopyAnnotations` should apply to the same constructor. So in other words I'm resisting the idea that `@CopyAnnotations` applied to a class would affect all `@AutoFactory` constructors within that class.

frankbenoit: @eamonnmcmanus 
Until now, I had the understanding, that `@CopyAnnotations` is always on the class, because it copies the class annotations onto the factory class.

With your statement, when it is placed onto the ctor (or multiple ctors), I (as a user) would interpret it as "copy ctor annotations to the create method".

What shall be the exact semantics and allowed constellations then, one of these?
1. If `@CopyAnnotations` is present on one or multiple of the ctors, it is like as it is on the class. The arguments must be exactly the same on each one. (redundancy)
2. `@CopyAnnotations` must be present either on none or all ctors, with all args exactly the same. (redundancy)
3. `@CopyAnnotations` must be present at most once on a ctors.
4. `@CopyAnnotations` is only allowed on class, and if present, the `@AutoFactory` is as well only allowed on the class

Or, we restrict `@CopyAnnotations` today to the class, accepting that it might be a bit confusing if the `@AutoFactory` is at some ctors and not at class.
This would allow in future the extension to allow `@CopyAnnotations` onto the ctor with the semantics to copy the ctor annotation onto the factory create method.

eamonnmcmanus: I think that `@CopyAnnotations` should be applied to the same thing (class or constructor) as `@AutoFactory`. I think it should not copy class annotations unless `fromClass` has a non-default value, and in that case it should obviously copy them from that class. It doesn't bother me that you might have to have _n_ very similar `@CopyAnnotations` if you have _n_ overloaded `@AutoFactory` constructors. I'm not sure that's very common. Searching over Google's internal codebase suggests that fewer than 1% of source files that contain `@AutoFactory` contain it more than once.

frankbenoit: @eamonnmcmanus 
the requirement is still not clear to me. Perhaps we can have a short talk, to save time? gmail user frank.rene.benoit

1. You say "should not copy class annotations unless fromClass has a non-default value". 
I thought this only switches to _another_ copy source. 
Do you say, it is mandatory to define fromClass to make it active at all?
Then the default value is useless, right?

```java
@AutoFactory
//@AutoFactory.CopyAnnotation // <- not sufficient
@AutoFactory.CopyAnnotation(fromClass = Foo.class)
@NonNullByDefault
class Foo {}
```

2. This shall give error, due to different copy source

```java
@NonNullByDefault
class Foo {
    @AutoFactory
    @AutoFactory.CopyAnnotation(fromClass = Foo.class)
    Foo(){}

    @AutoFactory
    @AutoFactory.CopyAnnotation(fromClass = Other.class)
    Foo(int arg){}
}
```

3. This is ok, only one CopyAnnotations needed.

```java
@NonNullByDefault
class Foo {
    @AutoFactory
    @AutoFactory.CopyAnnotation(fromClass = Foo.class)
    Foo(){}

    @AutoFactory
    Foo(int arg){}
}
```

eamonnmcmanus: Just as a reminder, we may end up with a different design after Google's internal API review process. I haven't started that process yet but intend to soon.

Nevertheless I think the exercise of implementing a proposal is a useful one because it is revealing problems. You have shown that it is probably a mistake to try to use the same annotation to control copying of class annotations and parameter annotations. I think it makes more sense to have separate `@AutoFactory.CopyClassAnnotations` and `@AutoFactory.CopyParameterAnnotations`. Then this PR would just be for the former. In that case it means that `@AutoFactory.CopyClassAnnotations` applies to the `@AutoFactory` class, whether or not the `@AutoFactory` annotation is applied to that class or to its constructor(s). There are a few things that are unclear, though:

1. Should it also be legal to apply `@AutoFactory.CopyClassAnnotations` to a constructor that has the `@AutoFactory` annotation? Otherwise, if it's applied to the class, it could be quite far away from the `@AutoFactory` annotation. But as you mention, there's then the issue of what happens when there is more than one `@AutoFactory` constructor.

2. Should it default to copying the annotations of the `@AutoFactory` class? The alternative is for `fromClass` not to have a default.

3. Would it be sufficiently clear if we renamed `fromClass` to `value`, so you could write `@AutoFactory.CopyAnnotations(SomeClass.class)` instead of `@AutoFactory.CopyAnnotations(fromClass = SomeClass.class)`?

I think you can make arbitrary choices for these questions, but we'll need to sort them out in the API review.

frankbenoit: @eamonnmcmanus can you tell, when this "API review" will take place? For me this issue is blocking to start using auto-factory in my work project.

frankbenoit: @eamonnmcmanus can you tell, when this "API review" will take place? For me this issue is blocking to start using auto-factory in my work project.

eamonnmcmanus: If you are blocked I would suggest forking AutoFactory until we are able to get to it.

eamonnmcmanus:
It has taken years, but we do now have a [reviewed API](https://github.com/google/auto/blob/d37e51da4d02e77355cba3d2f8d39b4e9a7d8cc8/factory/src/main/java/com/google/auto/factory/AutoFactory.java#L105) for this purpose. Thanks for this contribution, which was extremely helpful in getting the ball rolling, however slowly.

Issue #631, opened 2018-04-23
Create standalone annotations artifact for AutoFactory 


AngusMorton:
This was done for AutoValue (https://github.com/google/auto/issues/268) and should also be done for AutoFactory.

dbarwacz:
This seems like a trivial issue and triggers most of warnings on our builds, can I offer my help?

dbarwacz:
Also, it's a dupe of #632

filipkowicz:
> This was done for AutoValue (#268) and should also be done for AutoFactory.

it was done with https://github.com/JakeWharton/AutoValueAnnotations - it would be cool to have it here in original Api and it's seems to be quite easy as there is no need of change actual code, just generating and publishing artefacts - hope it will be available soon - especially now when gradle force you to declare annotationProcessors explicitly in gradle config

tbroyer:
@filipkowicz https://github.com/google/auto/releases/tag/auto-value-1.6

Wrt your Grade comment, for auto-service or auto-factory, just declare the same dependency in both `api` (or `compileOnly`) and `annotationProcessor` (the Android Gradle Plugin might have different requirements though). You can also exclude transitive dependencies (or add the dependency on the `@jar`) for the `api`/`compileOnly` configuration.

filipkowicz:
@tbroyer it's great but it's for `AutoValue` - I need same one` for AutoFactory`.

according to gradle: I'm currently doing it this way but:
- i need to exclude `Guava` package added by it (even if it's compileOnly it breaks `Dagger` file generation because `Dagge`r assume that `Guava` is in classpath)
- can't use `Kapt` (kotlin annotation processor) because if I include `AutoFactory` with `compileOnly` it force me to use `annotationProcessor` instead of `kapt`

so yes - currently it works but it could be a lot nicer and there is no point of adding classes from annotation processor to classpath during compile

saket:
Can we do this for AutoService as well?

technoir42:
> Wrt your Grade comment, for auto-service or auto-factory, just declare the same dependency in both api (or compileOnly) and annotationProcessor (the Android Gradle Plugin might have different requirements though). You can also exclude transitive dependencies (or add the dependency on the @jar) for the api/compileOnly configuration.

I updated to Android Studio 3.6 Canary 12 and there is a (new?) Lint check called `AnnotationProcessorOnCompilePath` which complains when I add AutoFactory to `compileOnly` configuration. I had to disable the check entirely.

```
dependencies {
[...]

PR #889, opened 2020-10-09
auto-factory-annoations artifact #631

frankbenoit:


frankbenoit:
proposal for #631 
I am not good in maven, please expect this not done correctly.

PR #983, opened 2021-02-07
Allow pass Comparator parameter to sorted collection builder

pawellabaj:
You can define a method foosBuilder(Comparator<T>) that returns the associated builder type for that sorted collection (ie. SortedSet or SortedMap) that uses given comparator to sort added values. In this example, we have an ImmutableSortedSet<String> which can be built using the countriesBuilder(Comparatror<String>) method:

```
@AutoValue
public abstract class Animal {
  public abstract ImmutableSortedSet<String> countries();

  public static Builder builder() {
    return new AutoValue_Animal.Builder();
  }

  @AutoValue.Builder
    public abstract ImmutableSortedSet.Builder<String> countriesBuilder(Comparator<String> comparator);
    public abstract Animal build();
  }
}
```

### DIFF
--- a/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueNotEclipseTest.java
+++ b/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueNotEclipseTest.java
@@ -16,18 +16,8 @@
 package com.google.auto.value;
 
 import static com.google.common.truth.Truth8.assertThat;
-import static java.util.Comparator.comparing;
-import static java.util.Comparator.naturalOrder;
-import static java.util.Comparator.reverseOrder;
 
-import com.google.common.collect.ImmutableSortedMap;
-import com.google.common.collect.ImmutableSortedSet;
-import com.google.common.truth.Truth;
-import java.util.Comparator;
 import java.util.Optional;
-import java.util.Set;
-import java.util.SortedMap;
-import java.util.SortedSet;
 import javax.annotation.Nullable;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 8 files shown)


eamonnmcmanus:
This is very nice, and obviously a good idea. Because of the way our development processes work, we'll probably end up making this change in Google's internal repo and then propagating it to GitHub as a separate PR that supersedes this one. That change will properly credit you. I expect that it will follow your code changes closely, apart from formatting adjustments and the like. The main differences I expect are:

1. I think we can make this more general than just `orderedBy(Comparator)`. We can say that if you have `Bar.Builder foosBuilder(Baz x)` then that works so long as there is either a `Bar.builder(Baz)` method or a `new Bar.Builder(Baz)` constructor. That's just a generalization of the existing no-argument [rule](https://github.com/google/auto/blob/master/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueTest.java). Since we can say `new ImmutableSortedSet.Builder<>(comparator)`, we don't need to introduce special knowledge of `orderedBy`.
2. `AutoValueCompilationTest` exists mostly for negative compilation tests. There are just a couple of cases where we compare the generated code against expected code, because that becomes annoying to maintain when the details of the generated code change. For a change like this, we'll typically just add one or two examples in [`AutoValueTest`](https://github.com/google/auto/blob/master/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueTest.java) and check that they compile and behave as expected. (I think we should try to make this more discoverable.) We'll probably want to check that `toBuilder()` works correctly in that test.

pawellabaj:
- [ ] Change the code to allow also using constructors
- [ ] move test to `AutoValueTest`

pawellabaj:
@eamonnmcmanus , 

I had problems with `com.google.auto.value.CompileWithEclipseTest`, that's why I put my unit test in `com.google.auto.value.AutoValueNotEclipseTest`.

Could you recommend some materials about how to deal with Eclipse compiler? How to get more meaningful error messages than just 
```
compileWithEclipseJava6(com.google.auto.value.CompileWithEclipseTest)  Time elapsed: 2.73 sec  <<< FAILURE!
Compilation should succeed
expected to be true
        at com.google.auto.value.CompileWithEclipseTest.compileWithEclipse(CompileWithEclipseTest.java:126)
        at com.google.auto.value.CompileWithEclipseTest.compileWithEclipseJava6(CompileWithEclipseTest.java:79)
```

eamonnmcmanus:
Usually you should see the Eclipse compiler output in the Maven test output? It's unfortunately very noise, because there are a lot of warnings, but if something's wrong then you should also see compile errors, or possibly an exception.

pawellabaj:
I've resolved problems with Eclipse compiler test

eamonnmcmanus:
During internal review, @cgdecker pointed out that someone could potentially call `builder.foosBuilder(comparator1)` and then `builder.foosBuilder(comparator2)` on the same builder. Since the property builder is cached, the second call would actually just return the same value as the first call. We think the simplest solution is probably to make it an error to call property builder more than once if it has a parameter. You don't need to implement that – I'll add it to your code in our internal version of this PR.

PR #1006, opened 2021-03-16
Sync "How do I..." between builders.md and builders-howto.md

perceptron8:
After reviewing #1003 (`Stepped` example is great BTW!), I've noticed that [builders.md](https://github.com/google/auto/blob/master/value/userguide/builders.md) and [builders-howto.md](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md) are out of sync.
This change will hopefully make missing builder patterns easier to find.


eamonnmcmanus (approved):
Thanks! Internally we have a script that keeps some other markdown files in sync. We should probably have it check these ones too.

PR #1067, opened 2021-04-13
Add method name to getterMustMatch error

ZekeSnider:
This adds the offending method name to the `getterMustMatch` error message. This makes it much less ambiguous to debug. e.g. `Method createdAt does not correspond to a property of MyBuilder`.

### DIFF
--- a/value/src/test/java/com/google/auto/value/processor/AutoBuilderCompilationTest.java
+++ b/value/src/test/java/com/google/auto/value/processor/AutoBuilderCompilationTest.java
@@ -647,8 +647,8 @@ public final class AutoBuilderCompilationTest {
     assertThat(compilation).failed();
     assertThat(compilation)
         .hadErrorContaining(
-            "[AutoBuilderBuilderWhatProp] Method three does not correspond to a parameter of Baz(int one,"
-                + " int two)")
+            "[AutoBuilderBuilderWhatProp] Method three does not correspond to "
+                + "a parameter of Baz(int one, int two)")
         .inFile(javaFileObject)
         .onLineContaining("three(int x)");
   }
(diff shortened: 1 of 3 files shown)


eamonnmcmanus:
Thanks! For the most part we haven't consistently included this sort of information in error messages because the error location is supposed to tell you which method it is. Are you using an environment where compilation errors don't quote the failing line for you? (I guess Maven typically is like that.)

ZekeSnider:
@eamonnmcmanus Yeah, we are using Bazel, which isn't printing the line context for the failure unfortunately.

eamonnmcmanus:
OK. We should probably revisit all the error messages with that in mind, but this PR is a good start.

eamonnmcmanus (approved):

value/src/test/java/com/google/auto/value/processor/AutoBuilderCompilationTest.java
```diff
@@ -647,7 +647,7 @@ public void alienOneArgMethod() {
     assertThat(compilation).failed();
     assertThat(compilation)
         .hadErrorContaining(
-            "[AutoBuilderBuilderWhatProp] Method does not correspond to a parameter of Baz(int one,"
+            "[AutoBuilderBuilderWhatProp] Method three does not correspond to a parameter of Baz(int one,"
```
netdpb, line 650: Can you please rewrap this so it doesn't go over 100 characters?

ZekeSnider: Yep, updated

netdpb (approved):

eamonnmcmanus (approved):

Issue #1075, opened 2021-04-16
Add tests for Gradle incremental compilation


eamonnmcmanus:
We [broke](https://github.com/google/auto/issues/1071) Gradle incremental compilation in AutoValue 1.8 and AutoFactory 1.0, and furthermore I had to make 3 attempts to fix it before succeeding. The problem is that we don't use Gradle internally and it isn't part of our tests. We should add a test that ensures that it doesn't break again.

@tbroyer [wrote](https://github.com/google/auto/pull/1073#issuecomment-821390274):
> You could have a look at https://github.com/tbroyer/gradle-incap-helper/tree/v0.3/integTest/src/test/kotlin/net/ltgt/gradle/incap/integTest, it uses Gradle Test Kit (so has to be run from a Gradle build) to drive another Gradle build and check its output (this is based on log messages, because it's not that easy to do it otherwise: tbroyer/gradle-incap-helper#2)
> Because this is testing net.ltgt.gradle.incap:incap-processor itself, the test code is itself an annotation processor; it would likely be simpler for the Auto projects.

Our tests are run from Maven, not Gradle, so it might be a bit trickier. We might need to launch one Gradle instance to have it run the test against a second Gradle instance.

tbroyer:
The Dagger build deploys the artifacts to the local repository to be able to easily load them from the Gradle builds (run separately from the GitHub Actions workflow).

eamonnmcmanus:
@tbroyer how does #1076 look to you?

Rishu-singhm:
That this code

PR #1076, opened 2021-04-18
Add a test for Gradle incremental compilation with AutoValue.

eamonnmcmanus:
Partial fix for #1075. Still need a test for AutoFactory.

Many thanks to @tbroyer for help developing this.

### DIFF
--- a/value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
+++ b/value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
@@ -131,7 +131,6 @@ public class GradleTest {
         return Optional.of(new File(gradleHome));
       }
     }
-    Path installationPath;
     try {
       Path gradleExecutable = Paths.get("/usr/bin/gradle");
       Path gradleLink =
@@ -139,7 +138,7 @@ public class GradleTest {
       if (!gradleLink.endsWith("bin/gradle")) {
         return Optional.empty();
       }
-      installationPath = gradleLink.getParent().getParent();
+      Path installationPath = gradleLink.getParent().getParent();
       if (!Files.isDirectory(installationPath)) {
         return Optional.empty();
       }
--- a/value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
+++ b/value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
@@ -1,7 +1,24 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.google.auto.value;
 
 import static com.google.common.truth.Truth.assertThat;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
+import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -64,22 +81,22 @@ public class GradleTest {
 
     // Set up the fake Gradle project.
     String buildGradleText = expandSystemProperties(BUILD_GRADLE_TEXT);
-    writeFile(fakeProject.newFile("build.gradle"), buildGradleText);
-    File srcDir = fakeProject.newFolder("src", "main", "java", "com", "example");
-    writeFile(new File(srcDir, "Foo.java"), FOO_TEXT);
+    writeFile(fakeProject.newFile("build.gradle").toPath(), buildGradleText);
+    Path srcDir = fakeProject.newFolder("src", "main", "java", "com", "example").toPath();
+    writeFile(srcDir.resolve("Foo.java"), FOO_TEXT);
 
     // Build it the first time.
     BuildResult result1 = buildFakeProject();
     assertThat(result1.getOutput()).contains(
         "Full recompilation is required because no incremental change information is available");
-    File output = new File(
-        fakeProject.getRoot(), "build/classes/java/main/com/example/AutoValue_Foo.class");
-    assertThat(output.exists()).isTrue();
+    Path output = fakeProject.getRoot().toPath()
+        .resolve("build/classes/java/main/com/example/AutoValue_Foo.class");
+    assertThat(Files.exists(output)).isTrue();
 
     // Add a source file to the project.
     String barText = FOO_TEXT.replace("Foo", "Bar");
-    File barFile = new File(srcDir, "Bar.java");
-    writeFile(new File(srcDir, "Bar.java"), barText);
+    Path barFile = srcDir.resolve("Bar.java");
+    writeFile(barFile, barText);
 
     // Build it a second time.
     BuildResult result2 = buildFakeProject();
@@ -89,7 +106,7 @@ public class GradleTest {
     // is not working, this will produce a message like this:
     //   Full recompilation is required because com.google.auto.value.processor.AutoValueProcessor
     //   is not incremental
-    assertThat(barFile.delete()).isTrue();
+    Files.delete(barFile);
     BuildResult result3 = buildFakeProject();
     assertThat(result3.getOutput()).doesNotContain("Full recompilation is required");
   }
@@ -117,11 +134,9 @@ public class GradleTest {
     Path installationPath;
     try {
       Path gradleExecutable = Paths.get("/usr/bin/gradle");
-      Path gradleLink = Files.readSymbolicLink(gradleExecutable);
-      if (!gradleLink.isAbsolute()) {
-        gradleLink = gradleExecutable.getParent().resolve(gradleLink);
-      }
-      if (!gradleLink.toString().endsWith("/bin/gradle")) {
+      Path gradleLink =
+          Files.readSymbolicLink(gradleExecutable).resolveSibling(gradleExecutable).normalize();
+      if (!gradleLink.endsWith("/bin/gradle")) {
         return Optional.empty();
       }
       installationPath = gradleLink.getParent().getParent();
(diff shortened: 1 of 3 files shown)


eamonnmcmanus:
> Is this what you're looking for? https://github.com/actions/virtual-environments/blob/4a2e5bca16d99dcf4f1ee58a4305bff7cf957fe5/images/linux/scripts/installers/java-tools.sh#L70

Thanks for the pointer! That tells me what I had already discovered, that the Gradle installation directory is `/usr/share/gradle-7.0`, and also that `GRADLE_HOME` is incorrectly set to just `/usr/share/gradle`. So my hacky code to resolve the `/usr/bin/gradle` symlink looks like it's needed, at least for now.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,171 @@
...
+          "plugins {",
+          "  id 'java-library'",
+          "}",
+          "repositories {",
+          "  maven { url = uri('${localRepository}') }",
+          "}",
+          "dependencies {",
+          "  compileOnlyApi      'com.google.auto.value:auto-value-annotations:${autoValueVersion}'",
+          "  annotationProcessor 'com.google.auto.value:auto-value:${autoValueVersion}'",
+          "  testImplementation  'junit:junit:${junitVersion}'",
```
tbroyer, line 39: Where is this value coming from? I don't see it anywhere in the repo. I suppose the build doesn't fail because the `test` task is skipped due to the absence of test sources, and possibly resolution of this dependency isn't even attempted as a result?

You could probably just remove that line, and possibly use `compileJava` or `classes` rather than `build` task in `buildFakeProject`.
```suggestion
```

tbroyer (approved):
I have comments but they are mostly nitpicking; this is good to go.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,171 @@
...
+          "    return new AutoValue_Foo(bar);",
+          "  }",
+          "}");
+
+  private static final Optional<File> GRADLE_INSTALLATION = getGradleInstallation();
+
+  @Test
+  public void basic() throws IOException {
+    String autoValueVersion = System.getProperty("autoValueVersion");
+    assertThat(autoValueVersion).isNotNull();
```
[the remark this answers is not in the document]
eamonnmcmanus: Well, I want the test to fail outright if the system property isn't set. It is in fact set, by pom.xml.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private BuildResult buildFakeProject() throws IOException {
+    GradleRunner runner = GradleRunner.create()
+        .withProjectDir(fakeProject.getRoot())
+        .withArguments("--info", "build");
+    if (GRADLE_INSTALLATION.isPresent()) {
+      runner.withGradleInstallation(GRADLE_INSTALLATION.get());
+    } else {
+      runner.withGradleVersion("7.0");
```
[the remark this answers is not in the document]
eamonnmcmanus: Thanks, going with `GradleVersion.current()`.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private static Optional<File> getGradleInstallation() {
+    String gradleHome = System.getenv("GRADLE_HOME");
+    if (gradleHome != null) {
+      File gradleHomeFile = new File(gradleHome);
+      if (gradleHomeFile.isDirectory()) {
+        return Optional.of(new File(gradleHome));
+      }
+    }
```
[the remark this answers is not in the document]
eamonnmcmanus: I'm mainly interested in running this test from GitHub Actions, and `GRADLE_HOME` is set there. Unfortunately it's [currently set to the wrong thing](https://github.com/actions/virtual-environments/issues/3221), but once that's fixed I may just require that either `GRADLE_HOME` be set or `/usr/share/gradle` exist. There's way too much logic here now for my liking.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
eamonnmcmanus: Thanks, that was left over from earlier experiments.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,171 @@
...
+    Optional<Path> coreJar;
+    Pattern corePattern = Pattern.compile("gradle-core-([0-9]+)\\..*\\.jar");
+    try (Stream<Path> files = Files.walk(installationPath.resolve("lib"))) {
+      coreJar =
+          files.filter(p -> {
+            Matcher matcher = corePattern.matcher(p.getFileName().toString());
+            if (matcher.matches()) {
+              int version = Integer.parseInt(matcher.group(1));
+              if (version >= 7) {
```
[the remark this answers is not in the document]
eamonnmcmanus: I'm fine with requiring 5 rather than 7, though again I'm hoping this logic will all disappear.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,170 @@
+package com.google.auto.value;
```
cgdecker, line 1: Needs a copyright header.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,170 @@
...
+      }
+    }
+    Path installationPath;
+    try {
+      Path gradleExecutable = Paths.get("/usr/bin/gradle");
+      Path gradleLink = Files.readSymbolicLink(gradleExecutable);
+      if (!gradleLink.isAbsolute()) {
+        gradleLink = gradleExecutable.getParent().resolve(gradleLink);
+      }
+      if (!gradleLink.toString().endsWith("/bin/gradle")) {
```
cgdecker, line 124: This could be `!gradleLink.endsWith("bin/gradle")`. Unlike a string comparison, that wouldn't be true if the string form ended in `"somethingbin/gradle"`.

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,170 @@
...
+  public void basic() throws IOException {
+    String autoValueVersion = System.getProperty("autoValueVersion");
+    assertThat(autoValueVersion).isNotNull();
+    String localRepository = System.getProperty("localRepository");
+    assertThat(localRepository).isNotNull();
+
+    // Set up the fake Gradle project.
+    String buildGradleText = expandSystemProperties(BUILD_GRADLE_TEXT);
+    writeFile(fakeProject.newFile("build.gradle"), buildGradleText);
+    File srcDir = fakeProject.newFolder("src", "main", "java", "com", "example");
```
cgdecker, line 68: This is really nitpicky, but could you use the `Path`-based APIs here where possible? I just don't like mixing usages of `File` and `Path` in one source file except for necessary conversion at API boundaries.

cgdecker (changes requested):

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,180 @@
...
+      File gradleHomeFile = new File(gradleHome);
+      if (gradleHomeFile.isDirectory()) {
+        return Optional.of(new File(gradleHome));
+      }
+    }
+    Path installationPath;
+    try {
+      Path gradleExecutable = Paths.get("/usr/bin/gradle");
+      Path gradleLink =
+          Files.readSymbolicLink(gradleExecutable).resolveSibling(gradleExecutable).normalize();
```
cgdecker, line 138: I think this is backwards? Should be `gradleExecutable.resolveSibling(Files.readSymbolicLink(gradleExecutable))` right?

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,180 @@
...
+      if (!gradleLink.endsWith("/bin/gradle")) {
```
cgdecker, line 139: I think this should be "bin/gradle", not "/bin/gradle", because "/bin/gradle" is an absolute path, and I don't think the way `endsWith` works when given an absolute path is what you want here.

cgdecker (changes requested):

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
cgdecker: Actually yeah, definitely backwards... we already know that `gradleExecutable` is absolute, so `gradleLink` is always just going to be `gradleExecutable` in this case.

eamonnmcmanus (commented):
I _think_ this is OK now. It certainly is tedious to test by pushing and waiting!

value/src/it/functional/src/test/java/com/google/auto/value/GradleTest.java
```diff
@@ -0,0 +1,180 @@
...
+      Path gradleLink =
+          gradleExecutable.resolveSibling(Files.readSymbolicLink(gradleExecutable));
+      if (!gradleLink.endsWith("bin/gradle")) {
+        return Optional.empty();
+      }
+      installationPath = gradleLink.getParent().getParent();
+      if (!Files.isDirectory(installationPath)) {
+        return Optional.empty();
+      }
+    } catch (IOException e) {
```
cgdecker, line 146: I think you should be able to combine this with the "catch (IOException e)" below (by moving the try-with-resouces block up into this try and not having it catch IOException itself). I think you could also move the declaration of installationPath inside the try block if you did that.

eamonnmcmanus: OK, I think this is what you had in mind?

cgdecker: Yep.

cgdecker (approved):

PR #1099, opened 2021-05-25
Implicitly exclude Kotlin `@Metadata` annotations from `@CopyAnnotations`

### LINKED ISSUE #1087: Feature request: implicitly exclude Kotlin `@Metadata` annotations from being copied
Context: kotlinc generates classes annotated with `@kotlin.Metadata` annotations on them. These annotations contain binary data about the kotlin language features, which kotlin then uses to understand those features in compiled binaries (jars, class files, etc).

These `@Metadata` annotations are not intended to be present on classes other than the one generated by kotlinc, let alone on Java classes. Doing so can confuse or break kotlin tooling and reflection that inspects these annotations. 

AutoValue will automatically copy these annotations though unless opted out via `@CopyAnnotations(excludes = [Metadata::class])`. I'd like to feature request _always_ excluding copying these, as it is always an error to copy them. Happy to send a PR if you're open to it!

ZacSweers:
Resolves #1087

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/AutoValueishProcessor.java
+++ b/value/src/main/java/com/google/auto/value/processor/AutoValueishProcessor.java
@@ -932,8 +932,25 @@ abstract class AutoValueishProcessor extends AbstractProcessor {
   ImmutableList<String> copiedClassAnnotations(TypeElement type) {
     // Only copy annotations from a class if it has @AutoValue.CopyAnnotations.
     if (hasAnnotationMirror(type, COPY_ANNOTATIONS_NAME)) {
-      Set<String> excludedAnnotations =
-          union(getExcludedAnnotationClassNames(type), getAnnotationsMarkedWithInherited(type));
+      Set<String> excludedAnnotations = ImmutableSet.<String>builder()
+          .addAll(
+              union(getExcludedAnnotationClassNames(type), getAnnotationsMarkedWithInherited(type)))
+          //
+          // Kotlin classes have an intrinsic @Metadata annotation generated
+          // onto them by kotlinc. This annotation is specific to the annotated
+          // class and should not be implicitly copied. Doing so can mislead
+          // static analysis or metaprogramming tooling that reads the data
+          // contained in these annotations.
+          //
+          // It may be surprising to see AutoValue classes written in Kotlin
+          // when they could be written as Kotlin data classes, but this can
+          // come up in cases where consumers rely on AutoValue features or
+          // extensions that are not available in data classes.
+          //
+          // See: https://github.com/google/auto/issues/1087
+          //
+          .add(ClassNames.KOTLIN_METADATA_NAME)
+          .build();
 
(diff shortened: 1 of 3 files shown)


eamonnmcmanus (approved):
Thanks!

value/src/main/java/com/google/auto/value/processor/AutoValueishProcessor.java
```diff
@@ -932,8 +932,25 @@ private boolean isInAutoValuePackage(String className) {
   ImmutableList<String> copiedClassAnnotations(TypeElement type) {
     // Only copy annotations from a class if it has @AutoValue.CopyAnnotations.
     if (hasAnnotationMirror(type, COPY_ANNOTATIONS_NAME)) {
-      Set<String> excludedAnnotations =
-          union(getExcludedAnnotationClassNames(type), getAnnotationsMarkedWithInherited(type));
+      Set<String> excludedAnnotations = ImmutableSet.<String>builder()
+          .addAll(
+              union(getExcludedAnnotationClassNames(type), getAnnotationsMarkedWithInherited(type)))
```
netdpb, line 937: You don't need the `union` anymore since you're using a builder. Just `addAll(getExcluded...).addAll(getAnnotations...)`.

value/src/test/java/com/google/auto/value/processor/AutoValueCompilationTest.java
```diff
@@ -3372,6 +3372,44 @@ public SourceVersion getSupportedSourceVersion() {
...
+    Compilation compilation =
+        javac()
+            .withProcessors(autoValueProcessor)
+            .withOptions("-Xlint:-processing", "-implicit:none")
+            .compile(test, metadata);
+    assertThat(compilation).succeededWithoutWarnings();
+    assertThat(compilation)
+        .generatedSourceFile("foo.bar.AutoValue_Test")
+        .contentsAsUtf8String()
+        .doesNotContain("@Metadata");
```
netdpb, line 3410: Just to be sure, should you look for "kotlin.Metadata" in case it's there but not imported? If you look for the full name without the `@`, that'll catch it whether it's imported or not.

netdpb (changes requested):

netdpb (approved):

eamonnmcmanus (approved):

ZacSweers:
@eamonnmcmanus @netdpb not to be that person in OSS, but is there anything else for autovalue I could help with toward another release? We're hoping to pick this up as it currently causes R8 to keep Metadata annotations unnecessarily, but we can also build from source if need be. Asking just in case I can help though!

eamonnmcmanus:
Oh I think we can easily make a new release to incorporate this. Stay tuned.

ZacSweers:
Thanks!

eamonnmcmanus:
OK, looks as if 1.8.2 is on Maven Central now.

ZacSweers:
Thanks again!

PR #1114, opened 2021-06-06
Add FIELD target to @Provided

SplotyCode:
This way we can just use Lombok to generate Constructors.
Example:
 ```
@AutoFactory
@RequiredArgsConstructor
public class Permission {
    @Provided
    private final UserCacheFactory userCacheFactory;
    private final String name;
    private final int access;
}
```
vs
```
@AutoFactory
public class Permission {
    private final UserCacheFactory userCacheFactory;
    private final String name;
    private final int access;
    public Permission(@Provided UserCacheFactory userCacheFactory, 
                      String name, int access) {
        this.userCacheFactory = userCacheFactory;
        this.name = name;
        this.access = access;
    }
}
```

### DIFF
--- a/factory/src/main/java/com/google/auto/factory/Provided.java
+++ b/factory/src/main/java/com/google/auto/factory/Provided.java
@@ -15,7 +15,7 @@
  */
 package com.google.auto.factory;
 
-import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.*;
 
 import java.lang.annotation.Target;
 
@@ -25,5 +25,5 @@ import java.lang.annotation.Target;
  *
  * @author Gregory Kick
  */
-@Target(PARAMETER)
+@Target(PARAMETER, FIELD)
 public @interface Provided {}

### CODE REVIEW

**Comment by SplotyCode** [comment]
@googlebot I signed it!

File: factory/src/main/java/com/google/auto/factory/Provided.java [hunk 1]
```diff
@@ -25,5 +25,5 @@
  *
  * @author Gregory Kick
  */
-@Target(PARAMETER)
+@Target(PARAMETER, FIELD)
```
eamonnmcmanus, line 28: This isn't syntactically correct.

eamonnmcmanus (commented):
It seems you haven't actually tried this, since the code in the PR doesn't compile. I'm wondering what the interaction will be between the code generated by AutoFactory and the modifications that Lombok then applies. I think we would probably want a test that shows that this does what you want it to. But before you write that test, I'm not sure we would want to make this change anyway. I think it is a bit strange to allow `@Provided` on a field just so that Lombok will copy it onto a constructor parameter. What does `@Provided` on a field mean for somebody who isn't using Lombok?

eamonnmcmanus:
We do encourage contributions to the Auto projects, but for feature requests it is probably better to create an issue first. We'll certainly want test coverage for a new feature, which means that the implementation is a nontrivial amount of work. It's best to know that the project maintainers agree with the proposed change before doing that work, and an issue allows us to have that discussion. I'm going to close this PR, but if you want to argue for making this change, please open an issue for that.

Issue #1189, opened 2021-11-01
NPE on missing service class


astubbs:
I think this is happening to me after changing moving my implementations around

```
java: FATAL ERROR: java.lang.NullPointerException: Cannot invoke 
"com.google.common.collect.ImmutableSet.stream()" 
because the return value of 
"javax.lang.model.element.AnnotationValue.accept(javax.lang.model.element.AnnotationValueVisitor, Object)"
 is null
  	at com.google.auto.service.processor.AutoServiceProcessor$1.lambda$visitArray$0(AutoServiceProcessor.java:306)
  	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:273)
  	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
  	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
  	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
  	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
  	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
  	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
  	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
  	at com.google.auto.service.processor.AutoServiceProcessor$1.visitArray(AutoServiceProcessor.java:307)
  	at com.google.auto.service.processor.AutoServiceProcessor$1.visitArray(AutoServiceProcessor.java:294)
  	at jdk.compiler/com.sun.tools.javac.code.Attribute$Array.accept(Attribute.java:349)
  	at com.google.auto.service.processor.AutoServiceProcessor.getValueFieldOfClasses(AutoServiceProcessor.java:293)
  	at com.google.auto.service.processor.AutoServiceProcessor.processAnnotations(AutoServiceProcessor.java:137)
  	at com.google.auto.service.processor.AutoServiceProcessor.processImpl(AutoServiceProcessor.java:121)
  	at com.google.auto.service.processor.AutoServiceProcessor.process(AutoServiceProcessor.java:109)
  	at org.jetbrains.jps.javac.APIWrappers$ProcessorWrapper.process(APIWrappers.java:152)
[...]

PR #1190, opened 2021-11-01
NPE on missing Service?

astubbs:
https://github.com/google/auto/issues/1189

Was trying to experiment against the CI, but it's disabled


eamonnmcmanus:
I enabled the CI, and indeed this does appear to reproduce the problem in #1189.

eamonnmcmanus:
Thanks, this is very helpful! We don't generally merge PRs here since the Source Of Truth is Google's internal repo, but I'll include this test as part of the fix.

eamonnmcmanus:
Superseded by https://github.com/google/auto/issues/1189. Thanks again!

PR #1200, opened 2021-11-13
Use `Objects.requireNonNull` wherever possible

reugn:
Generate cleaner, more concise sources.


reugn:
Please rescan for CLA.

value/src/main/java/com/google/auto/value/processor/autoannotation.vm
```diff
@@ -70,15 +70,11 @@ final class $className implements $annotationName, `java.io.Serializable` {
       $params[$p].type $members[$p] #if ($foreach.hasNext) , #end
 #end ) {
 #foreach ($p in $params.keySet())
-  #if (!$members[$p].kind.primitive)
```
eamonnmcmanus, line 73: I'm not sure I understand why the changes in this file need to be anything more than changing `if (x == null)`+`throw new NPE(msg)` into `Objects.requireNonNull(x, msg)`. And, in fact the opportunity to do so on line 76 seems to have been missed.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
```diff
@@ -75,16 +75,14 @@ ${modifiers}class $subclass$formalTypes extends $origClass$actualTypes {
     ## the constructor is called from the extension code.
 
     #if ($identifiers)
-    if ($p == null) {
-      throw new NullPointerException("Null $p.name");
-    }
+        this.$p = `java.util.Objects`.requireNonNull($p, "Null $p");
```
eamonnmcmanus, line 78: This change makes sense. However, the indentation convention in this file is that Velocity directives (like `#if`) and code lines are indented independently of each other. So this line should be indented the same as the `if` statement it replaces.

(It happens that the code indentation doesn't matter because a separate indentation pass will fix it up anyway, but I think following the conventions makes the code easier to follow.)

value/src/main/java/com/google/auto/value/processor/autovalue.vm
```diff
@@ -75,16 +75,14 @@ ${modifiers}class $subclass$formalTypes extends $origClass$actualTypes {
...
     #else
-    `java.util.Objects`.requireNonNull($p);
+        this.$p = `java.util.Objects`.requireNonNull($p);
     #end
 
-  #end
-
+  #else
```
eamonnmcmanus, line 83: Given that there is now an `#else`, I might be inclined to invert the sense of the `#if`:

```
#if ($p.kind.primitive || $p.nullable || ($builderTypeName != "" && $isFinal)
## ...comment...
this.$p = $p;
#elseif ($identifiers)
this.$p = `java.util.Objects`.requireNonNull($p, "Null $p");
#else
this.$p = `java.util.Objects`.requireNonNull($p);
#end
```

eamonnmcmanus (changes requested):
Thanks, this is helpful! In the Old Days we targeted Java 6 so we couldn't use `java.util.Objects`, but that hasn't been true for a while now, and we'd already started using it. We should certainly use it more consistently.

Using the value returned by `requireNonNull` leads to shorter code, though I think that doesn't really work out in `autoannotation.vm`.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
reugn: Fixed indentation. Please let me know if I missed something.

value/src/main/java/com/google/auto/value/processor/autovalue.vm
reugn: Yes, it looks better that way.

value/src/main/java/com/google/auto/value/processor/autoannotation.vm
reugn: Refactored.

eamonnmcmanus (dismissed):

eamonnmcmanus:
Just an update here: when running this change against Google's internal codebase we found a number of problems with Android tooling. I think they can probably be fixed, but it may be a while.

PR #1201, opened 2021-11-15
Set minimum Java version 1.7 in description

reugn:


value/annotations/pom.xml
```diff
@@ -24,17 +24,16 @@
...
   <name>AutoValue Annotations</name>
   <description>
-    Immutable value-type code generation for Java 1.6+.
+    Immutable value-type code generation for Java 1.7+.
   </description>
   <url>https://github.com/google/auto/tree/master/value</url>
 
   <properties>
-    <java.version>1.7</java.version>
+    <java.version>1.8</java.version>
```
reugn, line 36: https://github.com/google/auto/blob/master/value/userguide/index.md#what-java-versions-does-it-work-with

cgdecker: I don't think this change is accurate; that link says that the compiler must be a JDK 8+ compiler but that compilation can still be source/target 7. This would change the source/target versions, I believe.

reugn: @cgdecker you're right. Reverted.

cgdecker (approved):

PR #1290, opened 2022-03-25
MoreTypes fixes and improvements

AriaAdibi:
The branch contains three commits.

The first two, each fixes a small bug.
- Fixing the implementation of `equals()` of `ComparedElements`.
- The incorrect use of `instanceof`, which was replaced with `getKind()`.

The third one is my suggestion for increasing the readability of the code. Mainly, I find the use of the word _visiting_ confusing/incorrect.

The commit messages provide detailed explanations.

### DIFF
--- a/common/src/main/java/com/google/auto/common/MoreTypes.java
+++ b/common/src/main/java/com/google/auto/common/MoreTypes.java
@@ -17,12 +17,6 @@ package com.google.auto.common;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
-import static javax.lang.model.type.TypeKind.ARRAY;
-import static javax.lang.model.type.TypeKind.DECLARED;
-import static javax.lang.model.type.TypeKind.EXECUTABLE;
-import static javax.lang.model.type.TypeKind.INTERSECTION;
-import static javax.lang.model.type.TypeKind.TYPEVAR;
-import static javax.lang.model.type.TypeKind.WILDCARD;
 
 import com.google.common.base.Equivalence;
 import com.google.common.base.Optional;
@@ -101,8 +95,8 @@ public final class MoreTypes {
     return TypeEquivalence.INSTANCE;
   }
 
-  // So EQUAL_VISITOR can be a singleton, we maintain visiting state, in particular which types
-  // have been seen already, in this object.
+  // EQUAL_VISITOR can is a singleton. We maintain a visited state, which signifies what types
+  // have already been seen in this object.
   // The logic for handling recursive types like Comparable<T extends Comparable<T>> is very tricky.
   // If we're not careful we'll end up with an infinite recursion. So we record the types that
   // we've already seen during the recursion, and if we see the same pair of types again we just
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks. I agree that the `equals` implementation looks suspicious. I know that it's trying to do something very strange (as discussed in its comments), so there may be a reason that it can't do the normal thing. But even if so, that doesn't guarantee that the current code is correct, and it would clearly benefit from a comment even if it were :)

I notice that this passes the project's tests. I'll see what happens when I run more of the tests from across Google's codebase with it.

After that, we may try to figure out whether some of the style changes are worth separating into another PR. For example, I think that `visiting` may actually be correct: We appear to generate new sets as we go, so it might not be a complete set of everything we've ever visited. But I may have also an `add` call, so there may be yet more going on. Then there are smaller questions around static imports, `==` vs. `equals` for enums, braces around `if` bodies, the local variable that exists only so that we can apply `@SuppressWarnings`, etc.

eamonnmcmanus:
This is too big a change to review. Can you make a PR that contains just the functional changes that you think we should make? Then we can examine the stylistic changes separately, for example using early returns instead of big if/else statements.

It would also help if the code followed the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).

cpovirk:
[One of the commit messages](https://github.com/google/auto/pull/1290/commits/aa3a08513558fe22fdcff1caeb60d26951ee882a) mentions:

> The arguments of `that` object was not compared and was ignored.
> For arguments, only arguments of `this` was compared which was strange.

Referring to:

https://github.com/google/auto/blob/7569e39d6ead70deff6b9c315a7c1c2c29f195a5/common/src/main/java/com/google/auto/common/MoreTypes.java#L151-L155

Anyway, as for my testing, this PR doesn't change the result of any test in our depot. But I saw some comments about Eclipse in the code, so maybe we wouldn't expect anything unless we tested there?

cpovirk:
Oh, I meant to also observe that the fix looked like it might use TypeMirror.toString. I'm not sure if that method promises to return a "useful enough" string for our purposes, but it would seem nicer to avoid if practical. Plus, as it happens, someone just told us about a bug in that method :)

AriaAdibi:
Thank you for the review. My responses in order:

1. Response to: "there may be a reason `equals()` is implemented that way"
I don't believe that is the case. My understanding of the code is that it parallelly traverses through the pair of types, and when a comparison of the current elements (as explained in the code) is done it adds the already compared types to a Set. This is done to avoid the infinite loop explained in the source code.
When adding to the (hash)set, the hashing depends only on the elements. To reconcile the collisions the `equals` is needed and it needs to take into account the arguments of `that` element as well. The current version does not make any sense to me, The reason that it has passed the tests so far is that such an elaborate test does not exist.

2. visiting --> visited
As explained above, and by snippets like this
https://github.com/google/auto/blob/7569e39d6ead70deff6b9c315a7c1c2c29f195a5/common/src/main/java/com/google/auto/common/MoreTypes.java#L194-L202
It seems to me that maybe the author was not very proficient in English (other similar English mistakes are also present) and wrote *visiting* in error. As your comments also suggest this error may cause serious confusion, as the naming and the behavior (in my opinion) do not match. Hence, I propose these fixes.

3. Sure, I will create a pull request containing only the functional changes. However, I wanted to note that the readability changes are actually quite simple. 99% are just visiting --> visited, and the flattening of if structure. Nevertheless, this was merely a suggestion.

4. Regarding Eclipse, as explained I don't think the fix has anything to do with the Eclipse compiler. The problems with Eclipse have been addressed by the way the comparison and especially the "trick" to prevent infinite loops are written.

5. I believe the comparison using `toString()` is a practical one and a decent choice. Of course, I am open to alternative suggestions. My reasonings are:
* Note that the whole purpose of this portion of the source code is to have a mechanism that can compare types in the desired way (explained in the Javadoc). For this reason, using other available methods to compare, like `Types#isSameType()`, inside this mechanism defeats its whole purpose.
* Also since we traverse through the types and compare element by element, it is not necessary to take the type arguments into account.
Therefore, I believe the `toString()` comparison is a good choice.

6. I am not sure I understand. What bug and what method has already been reported?

Thank you.

cpovirk:
> 11\. I am not sure I understand. What bug and what method has already been reported?

Sorry: I mean that there's a bug in `TypeMirror.toString`: https://bugs.openjdk.java.net/browse/JDK-8281238

@eamonnmcmanus knows far more about this than I do, so he may be able to give you a response to all the information you supplied above. Thanks for #1294. I'll close this to continue the discussion there.

eamonnmcmanus:
Concerning "visiting" versus "visited", I originally wrote this code, am fairly proficient in English, and chose those words deliberately. The `visiting` set represents the comparisons we are _currently making_ in the path of recursive calls leading to the current point. Changing it to `visited` would suggest that it recorded comparisons we have _already made_, which is not the intent. But `visiting` isn't a great name either. Something like `comparing` or `currentlyComparing` would be better, I think. In the [cited](https://github.com/google/auto/pull/1290#issuecomment-1087125794) comment, writing "comparing" rather than "visiting" would help, and a few extra commas wouldn't hurt either.

PR #1297, opened 2022-04-06
Fix broken link

Bananeweizen:
This document and the target link are in the same folder. Currently the final link target contains duplicate "userguide" segment, leading to 404.


eamonnmcmanus:
Thanks! The way our source code is managed means that this PR won't be merged but will be superseded by an [equivalent one](https://github.com/google/auto/pull/1298) originating from our internal systems. You will nevertheless be credited. :-)

PR #1308, opened 2022-04-24
Small fixes for AnnotationValues

AriaAdibi:
1. First Commit
I believe the thrown exception type (`IllegalStateException`) present in the `defaultAction()` of `ArrayVisitor` is incorrect. For two reasons, first, I do not find it semantically sound, and second, based on the Javadocs of the methods using the `ArrayVisitor`. I have changed it to `IllegalArgumentException` type.

2. Second Commit
The `toString()` is missing from the `Equivalence` of `AnnotationValues`; for consistency, I have added it.

### DIFF
--- a/common/src/main/java/com/google/auto/common/AnnotationValues.java
+++ b/common/src/main/java/com/google/auto/common/AnnotationValues.java
@@ -152,6 +152,11 @@ public final class AnnotationValues {
               },
               null);
         }
+
+        @Override
+        public String toString() {
+          return "AnnotationValues.equivalence()";
+        }
       };
 
   /**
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/google/auto/common/AnnotationValues.java
```diff
@@ -348,7 +353,7 @@ public static char getChar(AnnotationValue value) {
 
     @Override
     public ImmutableList<T> defaultAction(Object o, Void unused) {
-      throw new IllegalStateException("Expected an array, got instead: " + o);
+      throw new IllegalArgumentException("Expected an array, got instead: " + o);
```
eamonnmcmanus, line 356: This seems correct. Technically, it is an incompatible change, though I think it's unlikely it would break production code. (It might conceivably break tests.) I'm running it past all of Google's internal tests just to check.

eamonnmcmanus: Tests all pass. We're good to go. Thanks!

eamonnmcmanus (approved):

PR #1309, opened 2022-04-24
Suggesting a rewording of `ElementName` Javadoc

AriaAdibi:
This is a proposed rewording of the Javadoc of `ElementName`, based on the quesion #1291.

### DIFF
--- a/common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
+++ b/common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
@@ -179,9 +179,7 @@ public abstract class BasicAnnotationProcessor extends AbstractProcessor {
   @Override
   public final ImmutableSet<String> getSupportedAnnotationTypes() {
     checkState(steps != null);
-    return steps.stream()
-        .flatMap(step -> step.annotations().stream())
-        .collect(toImmutableSet());
+    return steps.stream().flatMap(step -> step.annotations().stream()).collect(toImmutableSet());
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -538,9 +538,14 @@ private ElementName(Kind kind, Name name) {
     }
 
     /**
-     * An {@link ElementName} for an annotated element. If {@code element} is a package, uses the
-     * fully qualified name of the package. If it's a type, uses its fully qualified name.
-     * Otherwise, uses the fully-qualified name of the nearest enclosing type.
+     * An {@link ElementName} for an annotated element.
+     *
+     * <p>Packages declared and annotated in {@code package-info.java}, and type elements are tracked
```
eamonnmcmanus, line 543: I'm not sure I agree with this wording. If a package _isn't_ explicitly declared in `package-info.java` we will still track it directly. We won't track it via its nearest enclosing type, which doesn't exist. I think I would prefer to keep the existing text but also add another paragraph, like this:

> A package can be annotated if it has a `package-info.java` with annotations on the `package` declaration.

eamonnmcmanus: Hmm, if a package isn't declared in `package-info.java` then it can't be annotated, so this method won't be called. Still, I think I prefer my suggestion here.

AriaAdibi: Thank you for the feedback, changed it accordingly.

eamonnmcmanus (approved):

PR #1310, opened 2022-04-24
Fixed Readme of Auto and  Auto/Common

AriaAdibi:
Minor fixes for the Readme files of *Auto* and *Auto/Common*, the details are provided in the commit messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 A collection of source code generators for [Java][java].
 
-## Auto
+## Overview
 
 [Java][java] is full of code that is mechanical, repetitive, typically untested
 and sometimes the source of subtle bugs. _Sounds like a job for robots!_
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 A collection of source code generators for [Java][java].
 
-## Auto‽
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 
 A collection of source code generators for [Java][java].
 
-## Auto‽
```
eamonnmcmanus, line 7: I think this interrobang was deliberate. But it's also potentially confusing. If you just delete it then this section heading is the same as the previous one, which is not great. Maybe change it to something completely different, like `## Overview`.

common/README.md
```diff
@@ -14,7 +14,7 @@ annotation processing environment.
     and free from distortion from upstream compilation errors
 *   Visibility - utilities for working with Elements' visibility levels (public,
     protected, etc.)
-*   BasicAnnotationProcessor/ProcessingStep - simple types that
```
eamonnmcmanus, line 17: This change is definitely right. Thanks!

eamonnmcmanus (changes requested):

eamonnmcmanus (approved):

PR #1311, opened 2022-04-24
Enhanced `MoreTypesIsTypeOfTest`, and Fixed `MoreTypes.isType()` for primitives

AriaAdibi:
### First Commit, Enhanced `MoreTypesIsTypeOfTest`
The two main enhancements are:
1. The current approach for retrieving the desired `TypeMirror` is not ideal. Currently, an interface with an unused method for the sole purpose of extracting the return type of that method is defined, and then the type is extracted. I have used `Types` instead to get the desired `TypeMirror`.
2. The tests are the proper superset of previous tests; therefore, testing more scenarios. (and the code is more compact)

Additionally, I have a further suggestion, which I have not yet created a PR for, for extending `isTypeOf()` to work for hierarchical type, type arguments, and wildcards as well. If this PR is accepted, it would be easier for me to include my new suggestion unit tests later.

### Second Commit, Fixed `MoreTypes.isType()` for primitives
Please correct me if I am wrong. I believe that primitives cannot be referenced by `Class`; therefore, `isType()` should not return true for them. I have corrected this mistake.

### DIFF
--- a/common/src/main/java/com/google/auto/common/MoreTypes.java
+++ b/common/src/main/java/com/google/auto/common/MoreTypes.java
@@ -820,6 +820,11 @@ public final class MoreTypes {
     }
 
     @Override
+    public Boolean visitPrimitive(PrimitiveType type, Void p) {
+      return true;
+    }
+
+    @Override
     public Boolean visitArray(ArrayType array, Void p) {
       return true;
     }
--- a/common/src/main/java/com/google/auto/common/MoreTypes.java
+++ b/common/src/main/java/com/google/auto/common/MoreTypes.java
@@ -820,11 +820,6 @@ public final class MoreTypes {
     }
 
     @Override
-    public Boolean visitPrimitive(PrimitiveType type, Void p) {
-      return true;
-    }
-
-    @Override
     public Boolean visitArray(ArrayType array, Void p) {
       return true;
     }
--- a/common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
+++ b/common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
@@ -42,12 +42,12 @@ public class MoreTypesIsTypeOfTest {
 
   @Rule public CompilationRule compilationRule = new CompilationRule();
 
-  private Elements eltUtils;
+  private Elements elementUtils;
   private Types typeUtils;
 
   @Before
   public void setUp() {
-    this.eltUtils = compilationRule.getElements();
+    this.elementUtils = compilationRule.getElements();
     this.typeUtils = compilationRule.getTypes();
   }
 
@@ -179,8 +179,8 @@ public class MoreTypesIsTypeOfTest {
     }
   }
 
-  // Utility method(s) for this test.
+  /* Utility method(s) */
   private TypeElement getTypeElementFor(Class<?> clazz) {
-    return eltUtils.getTypeElement(clazz.getCanonicalName());
+    return elementUtils.getTypeElement(clazz.getCanonicalName());
   }
 }
--- a/common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
+++ b/common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
@@ -15,198 +15,171 @@
  */
 package com.google.auto.common;
 
-import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.google.common.truth.Truth.assertWithMessage;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-import com.google.common.collect.Iterables;
+import com.google.common.collect.ImmutableList;
 import com.google.testing.compile.CompilationRule;
-import javax.lang.model.element.Element;
+import java.util.List;
+import java.util.SortedMap;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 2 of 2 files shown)


common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
```diff
@@ -15,198 +15,171 @@
...
- * Tests {@link MoreTypes#isTypeOf}.
- */
+/** Tests {@link MoreTypes#isTypeOf(Class, TypeMirror)}. */
 @RunWith(JUnit4.class)
 public class MoreTypesIsTypeOfTest {
 
   @Rule public CompilationRule compilationRule = new CompilationRule();
 
-  private Elements elements;
+  private Elements eltUtils;
```
eamonnmcmanus, line 44: Would you mind spelling this as `elementUtils`? The abbreviation is a bit disconcerting.

common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
```diff
@@ -15,198 +15,171 @@
...
+            new PrimitiveTypeInfo(Byte.TYPE, Byte.class, TypeKind.BYTE),
+            new PrimitiveTypeInfo(Short.TYPE, Short.class, TypeKind.SHORT),
+            new PrimitiveTypeInfo(Integer.TYPE, Integer.class, TypeKind.INT),
+            new PrimitiveTypeInfo(Long.TYPE, Long.class, TypeKind.LONG),
+            new PrimitiveTypeInfo(Float.TYPE, Float.class, TypeKind.FLOAT),
+            new PrimitiveTypeInfo(Double.TYPE, Double.class, TypeKind.DOUBLE),
+            new PrimitiveTypeInfo(Boolean.TYPE, Boolean.class, TypeKind.BOOLEAN),
+            new PrimitiveTypeInfo(Character.TYPE, Character.class, TypeKind.CHAR));
+
+    for (int ki = 0; ki < 2; ki++) { // ki = 0: primitives, ki = 1: boxed primitives
```
eamonnmcmanus, line 77: I think it would be clearer to have something like:

```java
for (boolean boxedI : new boolean[] {false, true})
```
and likewise for `boxedJ` below.

common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
```diff
@@ -15,198 +15,171 @@
...
 
   @Test
   public void isTypeOf_fail() {
     assertFalse(
-        MoreTypes.isType(
-            getOnlyElement(typeElementFor(DeclaredVoid.class).getEnclosedElements()).asType()));
-    TypeMirror method =
-        getOnlyElement(typeElementFor(DeclaredVoid.class).getEnclosedElements()).asType();
+        MoreTypes.isType(getTypeElementFor(TestType.class).getEnclosedElements().get(0).asType()));
+    TypeMirror methodType = getTypeElementFor(TestType.class).getEnclosedElements().get(0).asType();
```
eamonnmcmanus, line 173: I think you could move this before the assertion and use it there? Also, `Iterables.getOnlyElement` might be a bit better than `.get(0)`.

common/src/main/java/com/google/auto/common/MoreTypes.java
```diff
@@ -819,11 +819,6 @@ public Boolean visitNoType(NoType noType, Void p) {
       return noType.getKind().equals(TypeKind.VOID);
     }
 
-    @Override
-    public Boolean visitPrimitive(PrimitiveType type, Void p) {
```
eamonnmcmanus, line 823: I think it might have made more sense for it to be like this from the start, but I am concerned that changing it now could break code depending on the previous behaviour. I would prefer to change the spec of `isType` to make it clearer exactly what it means when it says "a type that can be referenced by a `Class`". Primitives _can_ be referenced by a `Class`, for example `int.class`.

common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
```diff
@@ -15,198 +15,171 @@
...
-    TypeMirror type = extractReturnTypeFromHolder(typeElementFor(ArrayType.class));
-    assertWithMessage("array mirror represents an array Class object")
+    TypeMirror type = typeUtils.getArrayType(getTypeElementFor(String.class).asType());
+    assertWithMessage("Mirror:\t" + type + "\nClass:\t" + String[].class.getCanonicalName())
         .that(MoreTypes.isTypeOf(String[].class, type))
         .isTrue();
-  }
-
-  private interface PrimitiveBoolean {
-    boolean method();
```
eamonnmcmanus, line 78: I agree that getting rid of these is an improvement!

eamonnmcmanus (changes requested):

common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
AriaAdibi: No problem. Would you like me to revert it back to `elements`, as it was before?

Regarding the `eltUtils`, I just want to mention that this is the name used in the official java documents; for example, [here](https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Filer.html).

common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
AriaAdibi: Beautiful. :)

common/src/main/java/com/google/auto/common/MoreTypes.java
AriaAdibi: The change has been reverted.

common/src/test/java/com/google/auto/common/MoreTypesIsTypeOfTest.java
AriaAdibi: Thank you for your thorough answers, I really appreciate it. Done.

eamonnmcmanus (approved):

PR #1348, opened 2022-07-11
BasicAnnotationProcessor:: Changing deferral policy

AriaAdibi:
Following the discussion at [https://github.com/google/auto/discussions/1340](https://github.com/google/auto/discussions/1340), I believe changing the defer policy to individual rejected elements as opposed to their enclosing `TypeElement` (if not `PackageElement`) is a sound and beneficial choice for the following reasons. 

_For my convenience whenever I refer to an element, please assume that I am talking about a non-package element._

Before I explain I want to note that in my solution superficial validation for each element is still done starting from its enclosing `TypeElement`; therefore, the deferral policy when it comes to _well-informedness_ has not changed. Deferral policy is changed only for those elements that are rejected by a processor which are guaranteed to be well-informed.

### Is it backward compatible?
I believe it should be. Unless someone uses some hacks to overcome similar (rare) difficulties like that of the discussion.

### Why is it a sound choice?
1. The main problem in our discussion was that according to [process() in Processor Javadoc](https://docs.oracle.com/javase/8/docs/api/javax/annotation/processing/Processor.html) (_and only here_) the input is `TypeElement`s originating from the prior round and not any `Element`s. (Oddly enough, as it is verified in the discussion in practicality this is not the behavior of the checked Java compilers, more on this later.) I believe this is still the case especially if the current behavior of `BasicAnnotationProcessor` is acceptable.
 
    Presently, the aforementioned `process()` is overridden, and within it another `process()` method Is called. To distinguish between them let's call the latter `process2()`. The input of `process2()` is _not_ restricted to `TypeElement`s and in fact it gets any `Element` in general. With the current policy. the `step` deferred elements are only considered again in `process2()` and have nothing to do with `process()`. The only exception is when in `process()` the missing elements are to be reported, where now only enclosing types are reported but with my solution, the exact problematic elements are reported. I can only report the violating `TypeElement`s as well if required. For this reason, I believe the solution does not violate the aforementioned Javadoc. 

2. As hinted before, the checked processing tools (like OpenJDK) do not adhere to only sending the `TypeElement`s and in that case, the new solution is the more appropriate behavior to have.

### Why beneficial?
As discussed in the discussion because of the current behavior of common processing tools (like OpenJDK) if the annotation target something other than Types, it may happen that some elements correctly get processed and just because one other element within the same enclosing `TypeElement` gets rejected, all of these elements will be sent to be processed again. This is not optimized and also causes some headaches for the users (including me :)) of Google's `BasicAnnotationProcessor` to handle multiple processing scenarios.

### Asking for some favors :)
1. I would greatly appreciate it if you check the Javadoc of `ElementName` and let me know if I am correct and give me your opinion. I think the main purpose of `ElementName` instead of `Element` is the information presented here.
2. For the unit test I could not figure out the API to say more than 1 file is created with the processor. I would appreciate your help in this regard as well.

### Edits:
Now the retrieval of overloaded executable elements is done correctly. For now to gain access to `erasedParameterTypes()` I have added an instance of `ExplicitOverrides` in `ExecutableElementName`. While this is not bad it is subjective and for me is less than ideal, if this proposal gets accepted please let me know your opinion about a slight change in structure so that this method is more generally accessible.

I would still greatly appreciate your comments about the questions that I have asked before.

Thank you!


eamonnmcmanus:
Thanks for this! I'm finally getting around to it.

To answer some of the questions above:

* I think that rather than calling `ExplicitOverrides.erasedParameterTypes`, it might be better to do something like `MoreTypes.asExecutable(executableElement.asType()).getParameterTypes()` and then erase those yourself with `Types.erasure`.
* I'm not sure I understand the question about the unit test, but you should be able to assert about each generated file in its own statement. You should be using the more recent compile-testing API, though, like [this](https://github.com/google/auto/blob/de4299c1b717d878b0f13dd01f01997a9d2d3a32/value/src/test/java/com/google/auto/value/processor/AutoValueCompilationTest.java#L126).

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+                          isAcceptableElementKind(element)
+                              && simpleName.equals(element.getSimpleName())
+                              && hasSameErasedParametersType(
+                                  Objects.requireNonNull(
+                                      explicitOverrides.erasedParameterTypes(
+                                          (ExecutableElement) element, enclosingTypeElement))))
+                  .collect(MoreCollectors.onlyElement()));
+    }
+
+    private boolean hasSameErasedParametersType(
```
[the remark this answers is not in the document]
eamonnmcmanus: I realize that the fix I suggested isn't really enough. We would also have to handle the case where we have `Set<BarFactory>`, `Set<? extends BarFactory>`, even a `<T extends BarFactory>` type parameter on the method. That would suggest writing a `TypeVisitor` similar to [the one](https://github.com/google/auto/blob/01bf873bba2b23e6c07d9451da4fc4f832b87a43/common/src/main/java/com/google/auto/common/MoreTypes.java#L190) used by `MoreTypes.equivalence()`.

However!

I think there is an easier way. What we're trying to do is to find this method again in a later round after we saw it in an earlier one. And the issue is that there might be other methods with the same name but different parameter types. But we should be able to rely on the specification of [`TypeElement.getEnclosedElements()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.compiler/javax/lang/model/element/TypeElement.html#getEnclosedElements()), which essentially says that the elements will be returned in source order. That order shouldn't change between rounds, so you can just remember that this is the second method called `foo`, or even just the second element overall, to find it again. (That might fail if something like Lombok is hacking the AST, but I think that's an acceptable restriction.)

I'm going to see if I can make this idea work and run it past all of Google's compilations again.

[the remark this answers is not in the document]
eamonnmcmanus: Actually, we're looking at erased types, aren't we? So none of the cases I mentioned applies, though arrays do. Nevertheless I think this idea is probably cleaner than comparing the `getSimpleName()` of a `TypeElement` against the `getSimpleName()` of `typeUtils.asElement(errorType)`.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -103,20 +108,23 @@
  * body), then the well-formedness check will not defer processing {@code A}, but a processing step
  * can reject {@code A}.
  */
+@SuppressWarnings("Guava")
```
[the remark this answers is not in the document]
AriaAdibi: Noted, I will add a description for `@SuppressWarnings` from now on. I removed this one.

Correct Guava's `Optional` was the main reason. If I could be of any assistance for `Optional` or any other changes please don't hesitate to inform me. I would be glad to help.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -225,22 +233,25 @@ private void process(ImmutableSetMultimap<TypeElement, Element> validElements) {
         Set<? extends Element> rejectedElements =
             step.process(toClassNameKeyedMultimap(stepElements));
         elementsDeferredBySteps.replaceValues(
-            step, transform(rejectedElements, ElementName::forAnnotatedElement));
+            step,
+            transform(
+                rejectedElements,
+                (element) -> ElementName.forAnnotatedElement(element, typeUtils)));
```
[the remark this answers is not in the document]
AriaAdibi: Done. I also used `stream` API instead of `transform`.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -288,12 +299,17 @@ private ImmutableSetMultimap<TypeElement, Element> validElements(RoundEnvironmen
       Set<? extends Element> roundElements = roundEnv.getElementsAnnotatedWith(annotationType);
       ImmutableSet<Element> prevRoundElements = deferredElementsByAnnotation.get(annotationType);
       for (Element element : Sets.union(roundElements, prevRoundElements)) {
-        ElementName elementName = ElementName.forAnnotatedElement(element);
+        ElementName elementName = ElementName.forAnnotatedElement(element, typeUtils);
         boolean isValidElement =
             validElementNames.contains(elementName)
                 || (!deferredElementNames.contains(elementName)
+                    // For every element that is not module/package, to be well-informed its
```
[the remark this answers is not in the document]
AriaAdibi: Done. The position is also changed.
The reason I positioned it there was that it justifies the parameters of `validateElement()`, not `isValidElement` variable. I personally like my positioning better for this reason. Please let me know if you agree and if so I can change it back.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
   }
 
   /**
-   * A package or type name.
+   * An {@link ElementName} for an annotated element.
    *
-   * <p>It's unfortunate that we have to track types and packages separately, but since there are
-   * two different methods to look them up in {@link Elements}, we end up with a lot of parallel
-   * logic. :(
+   * <p>Instead of saving elements, an {@code ElementName} is saved since
```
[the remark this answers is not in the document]
AriaAdibi: I greatly appreciate your comment and insight into this. The changes have been made.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+
     /**
-     * An {@link ElementName} for an annotated element. If {@code element} is a package, uses the
-     * fully qualified name of the package. If it's a type, uses its fully qualified name.
-     * Otherwise, uses the fully-qualified name of the nearest enclosing type.
-     *
-     * <p>A package can be annotated if it has a {@code package-info.java} with annotations on the
-     * package declaration.
+     * Returns the {@link Element} corresponding to the name information saved in {@link
+     * ElementName}. {@link Optional#absent()} if non exists.
```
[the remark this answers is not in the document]
AriaAdibi: Great observation. Thanks.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+        return false;
+      }
+
+      PackageElementName that = (PackageElementName) object;
+      return this.qualifiedName.equals(that.qualifiedName);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(qualifiedName);
```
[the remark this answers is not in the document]
AriaAdibi: Done.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+      this.qualifiedName = MoreElements.asType(element).getQualifiedName();
+    }
+
+    @Override
+    protected boolean isAcceptableElementKind(Element element) {
+      return MoreElements.isType(element);
+    }
+
+    @Override
+    Optional<TypeElement> getElement(Elements eltUtils) {
```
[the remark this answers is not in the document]
AriaAdibi: Great observation. I missed this. Thanks.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -16,18 +16,16 @@
...
-import static com.google.auto.common.MoreElements.asPackage;
 import static com.google.auto.common.MoreStreams.toImmutableMap;
 import static com.google.auto.common.MoreStreams.toImmutableSet;
 import static com.google.auto.common.SuperficialValidation.validateElement;
-import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.Iterables.transform;
 import static com.google.common.collect.Multimaps.filterKeys;
 import static java.util.Objects.requireNonNull;
-import static javax.lang.model.element.ElementKind.PACKAGE;
```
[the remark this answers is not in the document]
AriaAdibi: I reverted it.

I believe I have a bit verbose style; however, I am modifying my style as time goes by. Accordingly, I appreciate feedbacks like this. Thank you!

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+   *   <li>and possibly other issues (such as reconstruction of AST after each round, and direct
+   *       manipulations of AST by tools like lombok) may make the saved element reference
+   *       unreliable.
+   * </ol>
    */
-  private static final class ElementName {
-    private enum Kind {
-      PACKAGE_NAME,
-      TYPE_NAME,
+  private abstract static class ElementName {
```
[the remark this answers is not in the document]
AriaAdibi: I agree. I tried to trim it as much as I could; however, unfortunately, the new fixes made changes big again. I will elaborate more on other comments. Please let me know if you have any ideas.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+
+    private ElementName(Element element) {
+      Objects.requireNonNull(element);
+      if (!isAcceptableElementKind(element)) {
```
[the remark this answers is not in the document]
AriaAdibi: I have removed the check; however, not the `isAcceptableElementKind()` method because I believe it makes the code more symmetric and readable. Please let me know what you think.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+        case FIELD:
+        case ENUM_CONSTANT:
+          return new FieldElementName(element);
+        case CONSTRUCTOR:
+        case METHOD:
+          return new ExecutableElementName(element, typeUtils);
+        case PARAMETER:
+          return new ParameterElementName(element, typeUtils);
+        default:
+          throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
AriaAdibi: Correct. The reason I did not include `TYPE_PARAMETER` was that I saw this `ElementKind` being ignored in `findAnnotatedElements()`. I have added the support for it, please let me know what you think.

As for `RECORD` and `RECORD_COMPONENT`, I have not thought about them; thank you! With the current solution, I just used `element.getKind().name()` for all to be symmetric and did not treat types differently as you suggested. Please let me know what you think here.

I am not sure why you mentioned using reflection to call `getAccessor()` could you please elaborate on that? The current solution does not use it.

That you for the test template, I added one unit-test using it.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+     * Returns the {@link Element} corresponding to the name information saved in {@link
+     * ElementName}. {@link Optional#absent()} if non exists.
      */
-    static ElementName forAnnotatedElement(Element element) {
-      return element.getKind() == PACKAGE
-          ? new ElementName(Kind.PACKAGE_NAME, asPackage(element).getQualifiedName())
-          : new ElementName(Kind.TYPE_NAME, getEnclosingType(element).getQualifiedName());
+    abstract Optional<? extends Element> getElement(Elements eltUtils);
+
+    abstract String getName();
```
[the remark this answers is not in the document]
AriaAdibi: `getName()` is removed. 

For `toString()` and `hashCode()` I tried to do this but some rare cases cause me to overwrite them again. I could not think of a better way. Please see the code and let me know if I have missed an idea to simplify this. Thanks.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -511,73 +521,372 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+    @Override
+    protected boolean isAcceptableElementKind(Element element) {
+      return element.getKind().isField();
+    }
+
+    @Override
+    Optional<VariableElement> getElement(Elements eltUtils) {
+      Optional<TypeElement> optionalEnclosingTypeElement =
+          enclosingTypeElementName.getElement(eltUtils);
+      if (!optionalEnclosingTypeElement.isPresent()) {
```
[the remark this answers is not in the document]
AriaAdibi: Done.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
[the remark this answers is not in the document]
AriaAdibi: Nicely spotted, thank you!

I have implemented this lookup using your suggestion.

I need your opinion regarding one slight problem that I have found.
Interestingly, I found that `Types#isSameType()` returns true for the `ERROR` type regardless of the other type. In hindsight this makes sense. One thing I have noticed about (open-)JDK behavior is the following which is peculiar and is a slight problem for our efforts. I believe this slight problem is the result of the aforementioned fact. Let me explain using an example:
``` java
@Annotation
<C extends Set<String>> void overloadedMethod(C x) {}
@Annotation
<C extends SortedSet<String>> void overloadedMethod(C x) {}
@Annotation
void overloadedMethod(SomeToBeGeneratedClass x) {}
@Annotation
<T extends TreeSet<String>> void overloadedMethod(T x) {}
```

When javac is scanning for annotated elements, and `SomeToBeGeneratedClass` references an `ERROR` type, then the last overloaded method is not reported by `RoundEnvironment#getElementsAnnotatedWith()`. Consequently, it will never get processed. It may be beneficial to note that the first two methods, although having the same `toString`,  are reported. Moreover, if the last method position moves above the `ERROR` referencing method, then in that case it will be reported. This feels like a bug to me, and if indeed it is a bug it would be great if I be able to contribute to open-JDK and fix it. Do you think this is an open-JDK bug? In the meantime, I believe this problem can only occur for scenarios exactly like this one. I can put simple check for it and simply handle it. Please let me know what you think. I will push this simple check in the near future.

AriaAdibi:
Thank you for your thorough comments. Sorry for the delayed response, I have a relatively busy schedule.

I have made the changes requested and responded as always I would appreciate your comments. I have also made additional changes and I want to mention a few of them and provide the reason(s). The change has got a little big and therefore, I have tried to organize my commits relatively isolated and meaningful to help the reviewer(s) to see the changes better.

1. I have added support for `TypeParameter` which was missing. I am not sure why before, `TypeParameter` was completely ignored. For example, if `class A<@Annotation T>` gets deferred then `T` would not have been processed at all. (I believe I have noticed a bug in deferral when it comes to `TypeParameter` as well, but I will leave this topic for a later pull request.)
2. 8d05d4f1f69864e1ca39e2effeeeb454b648df2e is an unrelated fix. However, because the fix was very small I included it here. Please let me know if you want it in a separate pull request.
3. I believe `ElementBluePrint` is a more suitable name than `ElementName` therefore I have changed the name. I can revert the change if desired.
4. Please let me know your thoughts on `UnsupportedElementBluePrint`.
5. b1eb35efdb8f06ca809db711c527bb7df439c2b6 To support `RECORD_COMPONENT` I have implemented `getEnclosingType()` differently, and moved it to `MoreElements`.

Thank you.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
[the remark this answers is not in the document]
AriaAdibi: I did not modify the implementation of deferring the ill-informed elements sufficiently before. After noticing it, every problem is solved. I believe now the code is even more readable, I hope you find it the same.

AriaAdibi:
Hello @eamonnmcmanus,
I hope you are doing good.

The changes that you requested have been fulfilled, and I believe these changes are very beneficial. I have been away for a long time and did not send any reminders. I would appreciate it if you consider this PR again and let me know if you want anything modified.

I would respond promptly as I believe these changes are very beneficial. I personally see the benefit frequently, when I have annotation processors targeting only methods. 

I appreciate your time.
All the best.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +525,402 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
     }
   }
 
+  /* ********************************************************************************* */
+  /* Element blue prints ************************************************************* */
```
eamonnmcmanus, line 529: This is very minor, but the English word is "blueprint" with no space or hyphen. So the various identifiers should have `blueprint` or `Blueprint` in them.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -255,69 +271,96 @@ private String processingErrorMessage(String target) {
...
         ImmutableSetMultimap.builder();
 
-    Set<ElementName> validElementNames = new LinkedHashSet<>();
+    // For optimization purposes, the PackageOrTypeElementBluePrint that have already
+    // been verified to be well-formed are stored.
+    Set<PackageOrTypeElementBluePrint> wellFormedPackageOrTypeElementBluePrints =
+        new LinkedHashSet<>();
 
-    // Look at the elements we've found and the new elements from this round and validate them.
+    /* Look at the found ready deferred elements and the new elements from this round
```
eamonnmcmanus, line 313: I think this should be something like "the already-deferred elements"? But maybe I'm misunderstanding what "ready" means here.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +525,402 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+    private final TypeElementBluePrint enclosingTypeElementBluePrint;
+    private final Name simpleName;
+    private final int ordinalOverloadPosition;
+
+    private ExecutableElementBluePrint(Element element) {
+      super(element);
+      TypeElement enclosingTypeElement = getEnclosingType(element);
+      this.enclosingTypeElementBluePrint = new TypeElementBluePrint(enclosingTypeElement);
+      this.simpleName = element.getSimpleName();
+      int ordinalOverloadPosition = 1;
```
eamonnmcmanus, line 816: This might be a bit clearer?

```java
Set<? extends Element> enclosed = enclosingTypeElement.getEnclosedElements();
int ordinalOverloadPosition = 0;
for (int i = 0; i < enclosed.size() && !enclosed.get(i).equals(elt); i++) {
  if (elt.getSimpleName().equals(enclosed.get(i).getSimpleName() && ...) {
    ordinalOverloadPosition = i;
  }
}
```

Then it is 0-origin rather than 1-origin, which I think just means that `<` changes to `<=` in the for-loop below.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +525,402 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+      Iterator<? extends Element> iter = enclosingTypeElement.getEnclosedElements().iterator();
+      Element elt = iter.next(); // The collection is not empty
+      while (!element.equals(elt)) {
+        if (elt.getSimpleName().equals(element.getSimpleName()) && isAcceptableElementKind(elt)) {
```
eamonnmcmanus, line 820: I think this should maybe be `isAcceptableElementKind(element)`? Otherwise the check should be outside the loop.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +525,402 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
-   * A package or type name.
+   * An {@link ElementBluePrint} for an annotated element.
    *
-   * <p>It's unfortunate that we have to track types and packages separately, but since there are
+   * <p>Instead of saving elements, an {@code ElementBluePrint} is saved since there is no guarantee
+   * that any particular element will always be represented by the same object. (Reference: {@link
+   * Element}) For example, Eclipse compiler uses different Element instances per round.
+   */
+  private abstract static class ElementBluePrint {
+    public final String toString;
```
eamonnmcmanus, line 540: I think this could be `private`? Or at least not `public`.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +525,402 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
-      this.name = name.toString();
+    @Override
+    Optional<? extends ExecutableElement> getElement(Elements elementUtils) {
+      return enclosingTypeElementBluePrint
+          .getElement(elementUtils)
+          .map(
+              typeElement -> {
+                ListIterator<? extends Element> listIter =
+                    typeElement.getEnclosedElements().listIterator();
+                for (int i = 0; i < ordinalOverloadPosition; i++) { // ordinalOverloadPosition >= 1
```
eamonnmcmanus, line 841: Even though it's a bit less efficient, I think it would be worth using a stream to get a list of elements where `elt.getSimpleName().equals(simpleName) && isAcceptableElementKind(elt)`. Then I think you're just looking for `list.get(ordinalOverloadPosition)`, or maybe ±1.

common/src/main/java/com/google/auto/common/MoreElements.java
```diff
@@ -81,6 +81,22 @@ public PackageElement visitPackage(PackageElement e, Void ignore) {
     }
   }
 
+  /**
+   * Returns the nearest enclosing {@link TypeElement} to the current element, throwing an {@link
+   * IllegalArgumentException} if the provided {@link Element} is not enclosed by a type.
+   */
+  public static TypeElement getEnclosingType(Element element) {
```
eamonnmcmanus, line 88: Does this need to be public?

eamonnmcmanus (changes requested):
Thanks for picking this up again! I had forgotten how tricky it was.

I think we're on the right track here, but I had a couple of suggestions for simplifications. I'll also try running this against all of Google's internal tests to see if there are unexpected issues. (That won't be ready until tomorrow.)

eamonnmcmanus:
For the unit-test question, I see that the existing test is using the obsolete compile-testing API. I'll look into updating the test to use the newer API, which will unfortunately mean a little merging and rewriting on your part.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: Thank you for this.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: I've refined variable names and improved this comment for better readability and clarity of purpose.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: I think mine is correct. Could you please double-check? The `elt` is the changing variable.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: You are right. There is no need. Thank you for this.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: I have changed the implementation as requested.

common/src/main/java/com/google/auto/common/MoreElements.java
AriaAdibi: I believe so. Do you disagree? Note that I have put this method in `MoreElements` and I find it to be a helpful utility method.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: I've adjusted `ordinalOverloadPosition` to be 0-origin. It's worth noting that, in my implementation, this value represents the ordinal position solely among methods/constructors with the same name, not the order among all enclosed elements, as your code suggests.

While I'm confident there was a specific rationale for this choice, I currently don't recall the exact details. It's plausible that the decision was influenced by the uncertainty in the ordering of implicitly declared elements, such as the default constructor, as outlined in the Java documentation.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: I have changed `!element.equals(elt)` to `!elt.equals(element)` as I think that was the source of confusion.

eamonnmcmanus: Yeah, I was the one getting confused here.

common/src/main/java/com/google/auto/common/MoreElements.java
eamonnmcmanus: OK. Usually for a public API change like this we have an internal API review process. But I think this one is very straightforward and we can probably dispense with that.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
eamonnmcmanus: I see, yes. Although we could imagine that the different processing rounds could return `getEnclosedElements()` in a different order, I think that would be pretty surprising given the specification that you quote. So indeed just storing the position of the element among all enclosed elements might be sufficient? Without needing to have any specific logic for overloading, except perhaps a verification check to ensure that the simple name is still the one we expect.

Alternatively, if you want to be safe and keep the logic as it stands, I think it would be good to get rid of that `Iterator`. I always wince a bit when I see an `Iterator`. 🙂  What about just streaming over the enclosed elements here too, to produce a `List` of elements with the same name, and then using `indexOf` to get `ordinalOverloadPosition`. Perhaps rename that, too.

eamonnmcmanus (commented):
(By the way, I ran the previous version through Google's internal tests and everything passed.)

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: Honest, I feel safer having both name and position within rather than just relying on the ordinal position. This preference stems partly because of the implicit declarations and partly because of practical AST-changing tools like Lombok, as you mentioned before. Since the processing order isn't guaranteed, I feel more secure with my approach. However, if you think sticking to only the ordinal position is the way to go, I'm fine with that.

Regarding the Iterator, I get it. :) How about using a straightforward while loop?

Also, I could use some help with naming the ordinalOverloadPosition. I'm struggling to come up with a self-explanatory name. I changed it to ordinalPositionWithSameName, but I'm not convinced it's the best choice. If you have any suggestions, I'd love to hear them.

AriaAdibi:
"Hello @eamonnmcmanus,
I hope this message finds you well.

I wanted to gently remind you about the pending review, in case it got buried in the midst of your busy schedule.

Your attention to this matter is greatly appreciated.
Thank you!

common/src/main/java/com/google/auto/common/MoreElements.java
eamonnmcmanus: I'm still a bit unsure about this. It seems to me that if `element` is already a type, the name and specification of the method suggest that it should nevertheless return that type's enclosing type. But in fact the implementation returns `element` itself. I know the private method in `BasicAnnotationProcessor` was the same, but I think we should change either the behaviour or the name.

I think this would be a good candidate for a follow-on change, with this change retaining the private `getEnclosingType` for now.

eamonnmcmanus (changes requested):
Sorry I'm only finding time for this now!

I think at this point we can take these changes and put them through Google's internal review process. I'll probably make a few cosmetic tweaks, and the review may result in further changes. There'll be a separate PR, where you will be duly credited, which will close this one.

I left a comment about the new `getEnclosingType` method.

common/src/main/java/com/google/auto/common/MoreElements.java
eamonnmcmanus: (Also there should probably be an associated test.)

common/src/test/java/com/google/auto/common/BasicAnnotationProcessorTest.java
```diff
@@ -341,23 +437,63 @@ public void properlyDefersProcessing_argumentElement() {
...
-        new RequiresGeneratedCodeProcessor();
-    Compilation compilation =
-        javac()
-            .withProcessors(requiresGeneratedCodeProcessor, new GeneratesCodeProcessor())
-            .compile(classAFileObject, classBFileObject);
+    requiresGeneratedCodeDeferralTest(dependentTestFileObject, generatesCodeFileObject);
+  }
+
+  @Test
+  public void properlyDefersProcessing_recordComponent() {
```
eamonnmcmanus, line 452: FYI we're discovering that these tests aren't quite complete. This is testing that a record will be deferred until the next round if it contains a component whose type is not yet defined. However what's really new with this PR is the notion of rejecting elements in a round _even if they don't reference undefined types_. So I think there need to be test methods for that too. If I add them I discover some issues with the code, because the various `FooElementBlueprint` classes weren't actually being tested.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +536,391 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+    private final ElementKind elementKind;
+
+    private FieldOrRecordComponentElementBlueprint(Element element) {
+      super(element); // toString is its simple name.
+      this.enclosingTypeElementBlueprint = new TypeElementBlueprint(getEnclosingType(element));
+      this.elementKind = element.getKind();
+    }
+
+    @Override
+    Optional<VariableElement> getElement(Elements elementUtils) {
```
eamonnmcmanus, line 736: This isn't correct for a `RecordComponentElement`, which isn't a `VariableElement`.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -505,74 +536,391 @@ private ImmutableSetMultimap<Class<? extends Annotation>, Element> toClassKeyedM
...
+      return Optional.ofNullable(elementUtils.getTypeElement(toString));
+    }
+  }
+
+  private static final class TypeParameterElementBlueprint extends ElementBlueprint {
+    private final TypeElementBlueprint enclosingTypeElementBlueprint;
+
+    private TypeParameterElementBlueprint(Element element) {
+      super(element);
+      this.enclosingTypeElementBlueprint = new TypeElementBlueprint(element.getEnclosingElement());
```
eamonnmcmanus, line 686: This isn't correct for the type parameter of a method.

eamonnmcmanus (commented):
So you don't need to do anything, but just so you know we're discovering that there are some bugs that didn't show up because of gaps in the test coverage.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: Well spotted, thank you. I must admit at the time I was writing this in 2022, I just started working with Java 11 and did not know about records.

common/src/test/java/com/google/auto/common/BasicAnnotationProcessorTest.java
AriaAdibi: You are absolutely right, the testing is subpar. However, I am not sure I understand the potential problem here. I look forward to your tests. If I could do anything please don't hesitate to ask.

PR #1357, opened 2022-08-30
Common:: Minor Improvements

AriaAdibi:
### Three minor changes are proposed.
1. In `BasicAnnotationProcessor`, method `isAnnotationPresent()` is removed and `MoreElements.isAnnotationProcessor()` is used instead.
2. In `BasicAnnotationProcessorTest`, I find it more natural (as conveyed by the name as well) that `steps` have order. Consequently, I replace `ImmutableSet` with `ImmutableList` as the type of the `Iterable`.
3. In `Overrides`, `Element` kind is checked with `instanceof` operator; however, the official document states that it should not be done in this way. I used `getKind()` method instead.

### DIFF
--- a/common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
+++ b/common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
@@ -17,6 +17,7 @@ package com.google.auto.common;
 
 import static com.google.auto.common.MoreElements.asExecutable;
 import static com.google.auto.common.MoreElements.asPackage;
+import static com.google.auto.common.MoreElements.isAnnotationPresent;
 import static com.google.auto.common.MoreStreams.toImmutableMap;
 import static com.google.auto.common.MoreStreams.toImmutableSet;
 import static com.google.auto.common.SuperficialValidation.validateElement;
@@ -362,12 +363,6 @@ public abstract class BasicAnnotationProcessor extends AbstractProcessor {
         annotatedElements.put(annotationType, element);
       }
     }
-  }
-
-  private static boolean isAnnotationPresent(Element element, TypeElement annotationType) {
-    return element.getAnnotationMirrors().stream()
-        .anyMatch(
-            mirror -> MoreTypes.asTypeElement(mirror.getAnnotationType()).equals(annotationType));
   }
 
   /**
--- a/common/src/main/java/com/google/auto/common/Overrides.java
+++ b/common/src/main/java/com/google/auto/common/Overrides.java
@@ -25,6 +25,7 @@ import com.google.common.collect.Maps;
 import java.util.List;
 import java.util.Map;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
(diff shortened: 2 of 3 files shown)


eamonnmcmanus:
The `isAnnotationPresent()` change makes sense. Thanks!

For `ImmutableSet`, I am not as convinced. `ImmutableSet` does explicitly document that it is ordered, in the order elements were added when the set was built. And, we also have the set property here that there is no more than one occurrence of any given element.

AriaAdibi:
Thank you for your prompt reply.

As for the second suggestion, of course, it is subjective; however, if I may I want to ask one question and share my opinion.

In regards to order, I have never noticed the orderliness behavior of `ImmutableSet` thank you I learned something. :) Nevertheless, if retaining the order is desired I am personally not fond of relying on the underlying implementation to preserve the order. Furthermore, it might not be clear to new readers who are not too familiar with Guava as it was not to me that the order is preserved since that is not the inherent property of a set.

As for storing only distinct `step`s, I must confess I am still puzzled by the true intention of defining `steps` (especially why call it `steps`). I thought maybe one of the appeals is the ability to include a step multiple times in the steps. Vaguely, for example, if the first time it gets rejected the following steps are performed and then try on processing again or something similar.
If you have a good example in which the intent of having `steps` is demonstrated could you kindly please share it here? I personally used `steps` with two steps once just to have different deferral queues for two completely different processors which shared one annotation. I would appreciate your comment on this.

eamonnmcmanus:
Thanks for the excellent questions. I agree that the utility of steps is not very obvious. I looked at most of the nearly 100 subclasses of `BasicAnnotationProcessor` that we have at Google, and only about half a dozen had more than one step. All but one of those appeared to be similar to what you describe: two steps for two different annotations. Might as well just have two annotation processors then. The final one was doing something more complicated, but it wasn't clear to me that the step concept was particularly useful even there.

I still think that you are basically never going to want to have the same `Step` instance twice in the `Iterable` returned by the `steps()` method. However, your point about the name "step" is a good one: if these are steps, then we are doing them one after another, and that feels more like a list than a set.

eamonnmcmanus (approved):

AriaAdibi:
Thank you very much for your informative comment. I will revert the changes in regards to `List` conversion.

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
```diff
@@ -358,18 +358,12 @@ private static void findAnnotatedElements(
       default: // do nothing
     }
     for (TypeElement annotationType : annotationTypes) {
-      if (isAnnotationPresent(element, annotationType)) {
+      if (MoreElements.isAnnotationPresent(element, annotationType)) {
```
netdpb, line 361: I think this would be more readable if it were statically imported; the class name doesn't add any useful information. Other methods from `MoreElements` are already statically imported.

netdpb (changes requested):

common/src/main/java/com/google/auto/common/BasicAnnotationProcessor.java
AriaAdibi: Great observation. Done, thanks.

AriaAdibi:
I have added another minor fix. Its description is provided in the overview.

netdpb:
> I have added another minor fix. Its description is provided in the overview.

Looks good.

> Although I have both the `google-java-format` plugin and appropriate GoogleStyle some style changes have been done upon saving the file. If anything is wrong in this regard or you want me to forcefully not apply these changes please let me know. Thank you.

Please revert the formatting changes. If you'd like, you can run [`google-java-format-diff.py`](https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py) on the whole PR (diffing between your branch and the PR base), but I believe it should be a no-op.

eamonnmcmanus (approved):

netdpb (approved):

PR #1695, opened 2024-01-19
System independent newline

### LINKED ISSUE #1656: Annotation Processor output is not deterministic
Hi,

the AutoService annotation processor uses a `Writer` for creating the `META-INF/service/` files. Here the `writer.newLine()` method is used to create the newlines in the file (see https://github.com/google/auto/blob/83b16dded09d864fbe2161d30c6b27f13a0883c3/service/processor/src/main/java/com/google/auto/service/processor/ServicesFiles.java#L87).

Sadly the method uses the os specific char for newline and therefore produces **non-deterministic builds**. 
Changing that would be an easy fix and I would be happy to provide a PR if you are interested in that.

hendrikebbers:
Fixes #1656

### DIFF
--- a/service/processor/src/main/java/com/google/auto/service/processor/ServicesFiles.java
+++ b/service/processor/src/main/java/com/google/auto/service/processor/ServicesFiles.java
@@ -34,6 +34,8 @@ import java.util.Set;
 final class ServicesFiles {
   public static final String SERVICES_PATH = "META-INF/services";
 
+  public static final char NEWLINE = '\n';
+
   private ServicesFiles() {}
 
   /**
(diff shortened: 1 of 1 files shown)


service/processor/src/main/java/com/google/auto/service/processor/ServicesFiles.java
```diff
@@ -34,6 +34,8 @@
 final class ServicesFiles {
   public static final String SERVICES_PATH = "META-INF/services";
 
+  public static final char NEWLINE = '\n';
```
eamonnmcmanus, line 37: I'm not sure it's worth defining a constant for this. Maybe just do `writer.write('\n')` below?

hendrikebbers: will change :)

eamonnmcmanus (approved):

eamonnmcmanus:
Thanks! Because the original source code is in Google's internal source repo, this PR will be superseded by one from [Copybara](https://github.com/google/copybara). You will still be credited.

PR #1751, opened 2024-04-09
Updated equals check in equals?

### LINKED ISSUE #1633: Remove `o == this` check from generated `equals`?
It appears that most cases `o == this` check actually degrades performance because of branch prediction algorithms. It might be worth considering to remove it from auto-value templates. Sorry for linking to YT, but there's no article anywhere - at least I couldn't find any.

Please see https://www.youtube.com/watch?v=kuzjX_efuDs.

capak07:
- Resolves #1633 
- Updated `o==this` check to check for property count
- Updated README's

### DIFF
--- a/value/src/main/java/com/google/auto/value/processor/autoannotation.vm
+++ b/value/src/main/java/com/google/auto/value/processor/autoannotation.vm
@@ -214,8 +214,16 @@ final class $className implements $annotationName, `java.io.Serializable` {
 
   @`java.lang.Override`
   public boolean equals($equalsParameterType o) {
-    if (o == this) {
-      return true;
+    int objectPropertyCount = 0;
+    for ($m in $members) {
+      if (!$m.kind.primitive) {
+        objectPropertyCount++;
+      }
+    }
+
+    if (objectPropertyCount > 1) {
+       if (o == this) {
+        return true;
     }
     if (o instanceof $annotationName) {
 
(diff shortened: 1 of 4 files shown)


chaoren:
I don't think this actually addresses https://github.com/google/auto/issues/1633. In fact, this change would probably perform worse than the current implementation. Also, we haven't decided on a resolution to https://github.com/google/auto/issues/1633 yet.

value/src/main/java/com/google/auto/value/processor/autoannotation.vm
```diff
@@ -214,8 +214,16 @@ final class $className implements $annotationName, `java.io.Serializable` {
 
   @`java.lang.Override`
   public boolean equals($equalsParameterType o) {
-    if (o == this) {
-      return true;
+    int objectPropertyCount = 0;
```
eamonnmcmanus, line 217: This can be done at compile time, so it shouldn't be done at run time.

value/src/main/java/com/google/auto/value/processor/autoannotation.vm
```diff
@@ -214,8 +214,16 @@ final class $className implements $annotationName, `java.io.Serializable` {
...
+    for ($m in $members) {
+      if (!$m.kind.primitive) {
+        objectPropertyCount++;
+      }
+    }
+
+    if (objectPropertyCount > 1) {
+       if (o == this) {
```
eamonnmcmanus, line 225: Braces don't match, so apparently this PR has never been tested.

eamonnmcmanus (commented):
I think we would be open to a carefully-constructed PR that implements #1633. This PR in its current state doesn't achieve that, however.

PR #1754, opened 2024-04-15
Add missing import to example on README

vorburger:


service/README.md
```diff
@@ -20,6 +20,8 @@ Say you have:
 ```java
 package foo.bar;
 
+import com.google.auto.service.AutoService;
+
```
eamonnmcmanus, line 24: There shouldn't be a blank line here.

eamonnmcmanus (changes requested):
Thanks!

FWIW, it's simpler for Googlers to send changes internally. This would be `//third_party/java_src/auto/service/README.md`. But we can do this one here.

vorburger:
https://github.com/google/auto/commit/5201a7c2cddba9a44030251c60fafe47849e82a4

Issue #318, opened 2016-04-08
AutoFactory: Add a way to inject directly, without Provider


anuraaga:
Currently it's only possible for factory parameters to be injected as Providers. In certain DI frameworks (e.g., spring), Provider#get can be slow (lots of reflection) or even dangerous (dependency graph isn't validated until runtime).

Make sense to add this?
```java
/**
 * An annotation to be applied to parameters that should be provided by an
 * {@linkplain javax.inject.Inject injected} value in a generated factory.
* Unlike {@link Provided}, this does not use {@link Provider} injection.
 */
@Target(PARAMETER)
public @interface Injected {}
```

Could also be an option to Provided but naming seems like it'd be clunky.
```java
@Provided(direct=true) String hmm,
@Provided(noProvider=true) String maybe,
@Provided(provider=false) String err
```

anuraaga:
@gk5885 Hey - let me know if either of these approaches sounds ok, and I can send a PR. Going back to writing manual factories is painful ;)

gk5885:
I don't follow the logic here.  Are you suggesting that `Provider.get()` is somehow more expensive than having just injected the value?  I agree that it can be slow, but it doesn't make sense to me that that cost would be any greater than the cost that you would have incurred before creating the object when you inject the value directly.

anuraaga:
Provider.get needs to first figure out what object to return before then returning it, and depending on the DI framework the performance of this will vary. In dagger, injection is all resolved at compile time so indeed this has almost no cost, but with something like Spring, it happens at runtime, using slow reflection (stuff like resolving qualifer annotations). A CPU profile of a server using autofactory with spring on critical paths showed a lot of samples stuck in this reflection - migrating them to handwritten, directly injected factories saved about 2% CPU overall, which is pretty significant for just removing the words Provider and get.

To me, the autofactory behavior of using Providers was surprising too - if you were to handwrite a factory, wouldn't you inject the values directly in almost all cases? The complexity of supporting direct injection seems low enough.

gk5885:
> Provider.get needs to first figure out what object to return before then returning it, and depending on the DI framework the performance of this will vary.

Yes, but the same holds true for any injection.  The fact that you're doing it with a `Provider` wrapper doesn't change the fact that all injections require dependency resolution.

> In dagger, injection is all resolved at compile time so indeed this has almost no cost, but with something like Spring, it happens at runtime, using slow reflection (stuff like resolving qualifer annotations)

[...]

PR #1773, opened 2024-04-29
Adds support for @AutoFactory(inject=true) which injects types directly

leourbina:
This makes it so that AutoFactory can be used with frameworks such as Weld which relies on @Produces annotations instead of Provider<T>.

This addresses the concern in https://github.com/google/auto/issues/318.


netdpb:
Hi. It looks like we haven't accepted the need for this feature (#318), so this pull request is premature. I'm skeptical of the need for this feature given the conversation on the request.

I'm going to close this PR for now; feel free to reopen it if the request is approved.

Issue #597, opened 2018-02-03
Please publish an Automatic-Module-Name


io7m:
Hello.

When using `auto` to generate code for a modular Java 9 project, you'll get complaints from tools about depending on an automatically-generated module name.

Is there any chance you could add an `Automatic-Module-Name` entry in the jar manifest in order to provide a stable name upon which other Java 9 projects can depend? I'd suggest:

```
Automatic-Module-Name: com.google.auto.value
```

ronshapiro:
AutoValue is entirely source-level annotations - does that change the picture?

io7m:
Nope: You're still required to specify a `requires static` clause in a modular project in order to be able to use the annotations at compile time.

heruan:
I'm not able to require the automatic module with the name that should be derived from the JAR file, i.e. `auto.service`.

I checked with:

```
$ jar --describe-module --file=$HOME/.m2/repository/com/google/auto/service/auto-service/1.0-rc4/auto-service-1.0-rc4.jar
No module descriptor found. Derived automatic module.

auto.service@1.0-rc4 automatic
requires java.base mandated
provides javax.annotation.processing.Processor with com.google.auto.service.processor.autoserviceprocessor
contains com.google.auto.service
contains com.google.auto.service.processor
```

but no luck, with `requires auto.service` I always get:

> `auto.service` cannot be resolved as a module

What am I missing?

io7m:
@eamonnmcmanus Any movement on this? It's very much a harmless one-line addition to the jar manifest.

tbroyer:
> very much a harmless

See http://blog.joda.org/2018/03/jpms-negative-benefits.html

> To emphasise this, just adding something to the MANIFEST.MF file can have an effect on how the project is run and tested. You need to test on both the class-path and module-path.

TomRK1089:
Stephen Colebourne's opinion aside, Java 12 is out now and 13 is fast approaching. The module system debuted with Java 9. If you don't want to release a modular version of this library you don't have to, but for the rest of us that are trying to use the new module system (and leverage benefits like the stripped down runtimes produced by `jlink` this is a very straightforward change with a lot of upside.

What is the _actual_ practical concern? Is there a reason to believe that an `Automatic-Module-Name` entry following the reverse domain convention that the Maven dependency itself uses is not sufficient?

UkonnRa:
Any update? In Java 14, `  requires com.google.auto.service;` does not work.

```
wonderland-java ./gradlew build           

[...]

PR #1976, opened 2025-09-21
Add Java module system support with Multi-Release JAR

cowwoc:
Add module-info.java files to all modules and configure Multi-Release JAR compilation to support both Java 8 (without modules) and Java 9+ (with modules).

- Add module declarations for common, service, value, and factory modules
- Configure Maven compiler for dual compilation (release=8 and release=9)
- Update build order and versions for consistency
- Maintain full backward compatibility with Java 8 runtime

Enables module system benefits for Java 9+ users while preserving existing Java 8 compatibility through Multi-Release JAR mechanism.

### DIFF
--- a/value/processor/pom.xml
+++ b/value/processor/pom.xml
@@ -21,11 +21,11 @@
   <parent>
     <groupId>com.google.auto.value</groupId>
     <artifactId>auto-value-parent</artifactId>
-    <version>HEAD-SNAPSHOT</version>
+    <version>999.0.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
   </parent>
 
   <artifactId>auto-value</artifactId>
-  <version>HEAD-SNAPSHOT</version>
   <name>AutoValue Processor</name>
   <description>
     Immutable value-type code generation for Java 8+.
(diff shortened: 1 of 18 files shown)


cowwoc:
Fixes #597

eamonnmcmanus:
The CI is failing with this message:
> Error: The action actions/upload-artifact@v3 is not allowed in google/auto because all actions must be pinned to a full-length commit SHA.

cowwoc:
@eamonnmcmanus Fixed. Try again.

value/processor/pom.xml
```diff
@@ -40,15 +40,15 @@
   </scm>
 
   <properties>
-    <auto-service.version>1.1.1</auto-service.version>
+    <auto-service.version>999.0.0</auto-service.version>
```
cpovirk, line 43: It might just be that we need to change this one.

We do need to change _something_ so that the version starts with a number: https://github.com/google/truth/pull/1601#discussion_r2366370155

cowwoc: Is this property meant to point to the latest released version? If so, I will fix this now.

cpovirk: That is my recollection, at least.

(Since the latest released version isn't modularized, that may complicate matters.)

ailton345ai-star (changes requested):
Pg

PR #2123, opened 2026-08-22
Copy javax and jakarta @Qualifier onto generated AutoFactory params

### LINKED ISSUE #1884: [Regression] AutoFactory 1.1.0 no longer propagates @Qualifier
Upgrading to AutoFactory 1.1.0 seems to cause objects `@Provided` with `@Qualifiers` to no longer work. It appears that the qualifier is not propagated to the generated factory class, whereas with version 1.0.1, it was.

Long story short, the AutoFactory _used_ to propagate the qualifier to the generated factory class:
```
  @Inject
  public MyClassFactory(@MyQualifier Provider<Integer> valueProvider) {
    ...
  }
```

Now, the qualifier is no longer propagated, causing dagger to be unable to find the `@Provides` annotation:
`error: [Dagger/MissingBinding] java.lang.Integer cannot be provided without an @Inject constructor or an @Provides-annotated method.
`
```
  @Inject
  public MyClassFactory(Provider<Integer> valueProvider) {
    ...
  }
```

To fully reproduce the issue, [here](https://drive.google.com/file/d/1HU2aYEJorXCZ8a4dOhTU19rNb7J2e6h0/view?usp=drive_link) is a link to a sample app, but if that doesn't work, I'll also paste the applicable code below. Please let me know if you have any questions, and I appreciate someone looking into it!

`MainActivity.java`:
```
import android.app.Activity;
import android.os.Bundle;
import android.widget.TextView;

import javax.inject.Inject;

public class MainActivity extends Activity {

    @Inject MyClassFactory myClassFactory;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        AppComponent component =
            DaggerAppComponent.builder()
                              .myModule(new MyModule())
                              .build();
        component.inject(this);


        TextView textView = findViewById(R.id.text_view); // Make sure you have this TextView in your layout

        // Using generated factory to create an instance with a non-injected constructor
        MyClass myClass = myClassFactory.create("Hello from AutoFactory!");
        System.out.println(myClass.getMessage() + " Value: " + myClass.getValue());
        textView.setText(myClass.getMessage() + " Value: " + myClass.getValue());
    }
```

`activity_main.xml`:
```
<?xml version="1.0" encoding="utf-8"?>
<androidx.constraintlayout.widget.ConstraintLayout
    xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    xmlns:tools="http://schemas.android.com/tools"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    tools:context=".MainActivity">

    <TextView
        android:id="@+id/text_view"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:text="Hello World!"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent" />

</androidx.constraintlayout.widget.ConstraintLayout>
```

`AppComponent.java`:
```
import dagger.Component;

@Component(modules = {MyModule.class})
public interface AppComponent {
    void inject(MainActivity activity);
}
```

`MyClass.java`:
```
import androidx.annotation.NonNull;

import com.google.auto.factory.Provided;

import javax.inject.Inject;

@AutoFactory
public class MyClass {

    private final String message;
    private final int value;

    @Inject
    public MyClass(@NonNull String message,
            @Provided @MyQualifier int value) {
        this.message = message;
        this.value = value;
    }

    public String getMessage() {
        return message;
    }

    public int getValue() {
        return value;
    }
}
```

`MyModule.java`:
```
import dagger.Module;
import dagger.Provides;

@Module
public class MyModule {

    @Provides
    @MyQualifier
    int provideValue() {
        return 42;
    }
}
```

`MyQualifier.java`:
```
import java.lang.annotation.Documented;
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

import javax.inject.Qualifier;

/**
 * Qualifier to indicate that the object is the context of the favorites screen.
 */
@Qualifier
@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target(value = {ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER})
public @interface MyQualifier {
}
```

tanvir-ux:
## Summary
- AutoFactory 1.1.0 chooses `jakarta.inject` whenever it is on the classpath, and then treated an annotation as a qualifier only if it was meta-annotated with **that** API's `@Qualifier`.
- Dagger / existing code still commonly uses `javax.inject.Qualifier`. Those annotations were dropped from generated factory constructors (`Provider<Integer>` instead of `@MyQualifier Provider<Integer>`), which is [issue #1884](https://github.com/google/auto/issues/1884).
- Qualifier detection now accepts either `javax.inject.Qualifier` or `jakarta.inject.Qualifier`, independent of which package is used for generated `@Inject` / `Provider`.

Fixes #1884

## Test plan
- [x] `mvn test` in `factory/` — `AutoFactoryProcessorTest` 185 tests, 0 failures (new mixed javax/jakarta cases included)
- [x] `mvn verify` in `factory/` — unit tests + invoker IT `functional` passed
- [ ] Reviewer: compile an `@AutoFactory` class with `@Provided @MyQualifier` where `@MyQualifier` is `javax.inject.Qualifier` and both inject APIs are on the processor classpath; generated constructor should keep `@MyQualifier` on the `Provider` parameter

### DIFF
--- a/factory/src/main/java/com/google/auto/factory/processor/InjectApi.java
+++ b/factory/src/main/java/com/google/auto/factory/processor/InjectApi.java
@@ -18,12 +18,14 @@ package com.google.auto.factory.processor;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static java.util.stream.Collectors.joining;
 
+import com.google.auto.common.MoreElements;
 import com.google.auto.common.MoreTypes;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import java.util.AbstractMap.SimpleEntry;
+import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
@@ -41,6 +43,9 @@ abstract class InjectApi {
 
   private static final ImmutableList<String> PREFIXES_IN_ORDER =
       ImmutableList.of("jakarta.inject.", "javax.inject.");
+
+  private static final ImmutableSet<String> QUALIFIER_ANNOTATIONS =
+      ImmutableSet.of("javax.inject.Qualifier", "jakarta.inject.Qualifier");
 
   static InjectApi from(Elements elementUtils, @Nullable String apiPrefix) {
     ImmutableList<String> apiPackages =
@@ -63,6 +68,23 @@ abstract class InjectApi {
   boolean isProvider(TypeMirror type) {
     return type.getKind().equals(TypeKind.DECLARED)
         && MoreTypes.asTypeElement(type).equals(provider());
+  }
+
+  /**
+   * True if {@code annotation} is meta-annotated with {@code javax.inject.Qualifier} or {@code
+   * jakarta.inject.Qualifier}.
+   *
+   * <p>Both are recognized even when generated code uses only one inject package. Otherwise, if
+   * {@code jakarta.inject} is preferred for {@code Inject}/{@code Provider} types, a {@code
+   * javax.inject.Qualifier} on a {@code @Provided} parameter would be dropped from the generated
+   * factory constructor.
+   */
+  static boolean isQualifier(AnnotationMirror annotation) {
+    return MoreElements.asType(annotation.getAnnotationType().asElement())
+        .getAnnotationMirrors()
+        .stream()
+        .map(meta -> MoreElements.asType(meta.getAnnotationType().asElement()))
+        .anyMatch(type -> QUALIFIER_ANNOTATIONS.contains(type.getQualifiedName().toString()));
   }
 
   private static ImmutableMap<String, TypeElement> apiMap(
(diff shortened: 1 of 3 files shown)


chaoren:
The existing code in `InjectApi` already tries to handle `Qualifier`, but it looks like you ignored all of that and implemented something from scratch.

tanvir-ux:
Thanks, you're right — I had bypassed InjectApi's existing Qualifier handling. Detection now uses InjectApi's Qualifier types via `isAnnotationPresent` for both packages when they're on the classpath, instead of a separate check.

tanvir-ux:
@chaoren gentle ping — the follow-up commit reuses `InjectApi`'s Qualifier `TypeElement`s via `isAnnotationPresent` for both packages instead of a separate check. Happy to adjust further if anything still looks off; otherwise ready for another look when you have a moment.

PR #2135, opened 2026-09-11
Point AutoValue invoker at value/src/it for functional tests

### LINKED ISSUE #1558: AutoValue functional tests not being run by Maven
It looks like the AutoValue functional tests (in `src/it/functional`) are compiled but not run by Maven. This was probably broken by #1457. I don't see evidence that any Maven target runs the tests: not `verify`, not `install`, not `integration-test`.

tanvir-ux:
## Summary
- The `maven-invoker-plugin` on `value/processor` defaulted to `processor/src/it`, so the functional IT projects under `value/src/it` (and the eclipse profile pass from `invoker.properties`) never ran.
- Set `projectsDirectory` to `${project.parent.basedir}/src/it` so invoker picks them up again.

Fixes #1558

## Test plan
- [ ] `mvn -pl value/processor -am verify` (or the repo `build-pom.xml` verify job) runs the functional invoker projects

### DIFF
--- a/value/processor/pom.xml
+++ b/value/processor/pom.xml
@@ -42,6 +42,8 @@
   <properties>
     <auto-service.version>1.1.1</auto-service.version>
     <errorprone.version>2.50.0</errorprone.version>
+    <!-- annotations is test-scoped on this module; invoker:install defaults to runtime. -->
+    <invoker.install.scope>test</invoker.install.scope>
   </properties>
 
   <dependencies>
@@ -208,8 +210,14 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-invoker-plugin</artifactId>
         <configuration>
-          <!-- ITs live under value/src/it, not processor/src/it (see #1558). -->
+          <!-- ITs live under value/src/it, not processor/src/it (see #1558).
+               Only functional is an invoker project (eclipse profile pass via
+               invoker.properties). gwtserializer is a reactor module under the
+               java-11-and-after profile, not an invoker IT. -->
           <projectsDirectory>${project.parent.basedir}/src/it</projectsDirectory>
+          <pomIncludes>
+            <pomInclude>functional/pom.xml</pomInclude>
+          </pomIncludes>
         </configuration>
       </plugin>
       <plugin>
--- a/value/processor/pom.xml
+++ b/value/processor/pom.xml
@@ -207,6 +207,10 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-invoker-plugin</artifactId>
+        <configuration>
+          <!-- ITs live under value/src/it, not processor/src/it (see #1558). -->
+          <projectsDirectory>${project.parent.basedir}/src/it</projectsDirectory>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/value/processor/pom.xml
+++ b/value/processor/pom.xml
@@ -42,8 +42,6 @@
   <properties>
     <auto-service.version>1.1.1</auto-service.version>
     <errorprone.version>2.50.0</errorprone.version>
-    <!-- annotations is test-scoped on this module; invoker:install defaults to runtime. -->
-    <invoker.install.scope>test</invoker.install.scope>
   </properties>
 
   <dependencies>
@@ -218,6 +216,12 @@
           <pomIncludes>
             <pomInclude>functional/pom.xml</pomInclude>
           </pomIncludes>
+          <!-- annotations is test-scoped here so default install.scope=runtime
+               would skip it; do not raise install.scope to test (pulls
+               compile-testing -> com.sun:tools, missing from Central). -->
+          <extraArtifacts>
+            <extraArtifact>com.google.auto.value:auto-value-annotations:${project.version}</extraArtifact>
+          </extraArtifacts>
         </configuration>
       </plugin>
       <plugin>
--- a/value/src/it/functional/pom.xml
+++ b/value/src/it/functional/pom.xml
@@ -182,11 +182,29 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <version>3.5.6</version>
+        <!-- gradle-test-kit pulls JUnit Jupiter onto the test classpath; Surefire then
+             auto-selects JUnitPlatformProvider and reports Tests run: 0 for these JUnit 4
+             tests (AutoValueTest et al). Pin the JUnit 4 provider. See #1558. -->
+        <dependencies>
+          <dependency>
+            <groupId>org.apache.maven.surefire</groupId>
+            <artifactId>surefire-junit4</artifactId>
+            <version>3.5.6</version>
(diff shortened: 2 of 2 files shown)


cpovirk:
The Auto build has always been too much of a mystery to me. I kicked off CI, which shows:

```
[INFO] [ERROR] Failed to execute goal on project gwtserializer: Could not resolve dependencies for project com.google.auto.value.it.gwtserializer:gwtserializer:jar:HEAD-SNAPSHOT
[INFO] [ERROR] dependency: com.google.auto.value:auto-value-annotations:jar:HEAD-SNAPSHOT (compile)
[INFO] [ERROR] 	Could not find artifact com.google.auto.value:auto-value-annotations:jar:HEAD-SNAPSHOT
```

tanvir-ux:
@cpovirk yeah — pointing at `value/src/it` made invoker pick up **gwtserializer** too (that’s a JDK11+ reactor module, not an invoker IT). Also `invoker:install` defaults to runtime scope, so test-scoped `auto-value-annotations` never got staged into `it-repo`.

Fix in `value/processor/pom.xml`:
- `pomIncludes` → only `functional/pom.xml`
- `invoker.install.scope=test` so annotations lands in the IT repo

Pushed; CI should re-run.

tanvir-ux:
(also: the new CI run is sitting on `action_required` for the fork — same as before, needs a maintainer approve/kick)

cpovirk:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-invoker-plugin:3.10.1:install (integration-test) on project auto-value: The following artifacts could not be resolved: com.sun:tools:pom:1.8.0_504 (absent): Could not find artifact com.sun:tools:pom:1.8.0_504 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
```

tanvir-ux:
@cpovirk thanks — that was from `invoker.install.scope=test`, which also tried to install `compile-testing`'s transitive `com.sun:tools` (the old JDK `tools.jar`), and that POM isn't on Central.

I dropped the broad test-scope install and stage only `auto-value-annotations` via `extraArtifacts` instead. Pushed in 29282c9. CI will need another maintainer kick/approve on the fork run when you have a moment.

cpovirk (dismissed):

eamonnmcmanus (changes requested):
Thanks for looking into this!

However, I'm not seeing any evidence from the logs that `value/` functional tests other than the gwtserializer ones are actually being run. For example [`AutoValueTest`](https://github.com/google/auto/blob/main/value/src/it/functional/src/test/java/com/google/auto/value/AutoValueTest.java). A good check would be to introduce a call to `fail()` in one of the test methods there. Then the [`mvn` command from the CI](https://github.com/google/auto/blob/3f75ef071d56609d358f9b42672cd553e79910bb/.github/workflows/ci.yml#L50) should fail.

cpovirk:
Sigh, sorry, thanks for checking that :(

tanvir-ux:
@eamonnmcmanus thanks — looking into this now. I'll verify that the `value/src/it/functional` suite (including `AutoValueTest`) actually runs under the CI `mvn` command, and push a fix if the invoker path still isn't picking it up.

tanvir-ux:
@eamonnmcmanus good catch — you were right that `AutoValueTest` wasn't actually executing.

CI on the previous HEAD was compiling `value/src/it/functional` (and the invoker clone) but Surefire reported **`Tests run: 0`**. Cause: `gradle-test-kit` puts JUnit Jupiter on the test classpath, so Surefire 3.x auto-selects `JUnitPlatformProvider` and skips the JUnit 4 suite (`AutoValueTest` et al.). `gwtserializer` doesn't pull Jupiter, which is why those tests still showed up.

Pushed a fix on this branch: pin `surefire-junit4` on the functional module (and the same provider on failsafe for `GradleIT`). After CI re-runs you should see `Running com.google.auto.value.AutoValueTest` in the logs. Fork workflows may need another approve/kick when you have a moment.

tbroyer:
Fwiw, Surefire 3.6 removes surefire-junit4 and always uses Jupiter (running Junit 4 tests with the Vintage engine)
https://maven.apache.org/surefire/maven-surefire-plugin/whats-new-3-6-0.html

So maybe it'd be best to first upgrade to 3.6 (#2134) and then reevaluate how to get the tests to run (if this doesn't directly fix it)

cpovirk:
(approving workflows again, though I am hoping to ultimately wait for https://github.com/google/auto/pull/2134 and then revert the final commit)

tanvir-ux:
@eamonnmcmanus @tbroyer follow-up: switched the functional IT to Surefire/Failsafe **3.6.0** plus `junit-vintage-engine` (and dropped the temporary junit4 provider pin).

That matches the Surefire 3.6 direction — Jupiter always, JUnit 4 via Vintage — so `AutoValueTest` should actually run once CI is kicked on the fork. Happy to adjust further if the logs still look wrong.

cpovirk:
Local testing showed:
- `TestEngine with ID 'junit-jupiter' failed to discover tests`
- or, if I remove the `junit-vintage-engine` configuration (leaving only the plugin bumps and the `maven-invoker-plugin` change), the problem solved by https://github.com/google/auto/pull/2140

So I pushed a merge commit. I'll trigger workflows now and see what happens, and I'll retest locally without the `junit-vintage-engine` configuration.

cpovirk:
CI shows the expected "failed to discover tests" error.

Local testing without the `junit-vintage-engine` configuration passes. I'm now checking that it fails when appropriate as [described above](https://github.com/google/auto/pull/2135#pullrequestreview-5180943155).

Does it seem reasonable that we'd be able to ultimately revert value/src/it/functional/pom.xml once https://github.com/google/auto/pull/2142 is incorporated?

tanvir-ux:
@cpovirk yes — that sounds right.

In fact I just dropped the `value/src/it/functional/pom.xml` Surefire 3.6 / `junit-vintage-engine` pin entirely (cb7c910). Your local result matches what CI showed: with Vintage, Surefire dies on `TestEngine with ID 'junit-jupiter' failed to discover tests`; without it (and with #2140 on main), the suite is fine again.

So this PR should only keep the invoker path fix in `value/processor/pom.xml`. Once #2142 lands, there's nothing left in `functional/pom.xml` from this PR to revert.

Happy to adjust further if the fail()-injection check still doesn't show `AutoValueTest` running.

cpovirk (approved):
If I put a failure into `com.google.auto.value.AutoValueTest.testEmpty` and include the Surefire bump to 3.6.0, I get a failing run. We'll get this merged after the bump, unless @eamonnmcmanus again pulls the trick of actually reviewing in more depth than I did and finds something wrong along the way :)

cpovirk:
Well, I had tested with the Failsafe version bump. I went back to test _without_ that just to be safe, and I got what I assume is a random flake:

```
[INFO] [INFO] #1 retrying connecting to the daemon 
[INFO] [INFO] #2 retrying connecting to the daemon 
[INFO] [INFO] #3 retrying connecting to the daemon 
[INFO] [ERROR] Failed connecting to the daemon in 4 retries
```

I'll retest to be _extra_ safe.

cpovirk:
Proper error confirmed.

